PR #19, opened 2026-01-12
Fix: Docker Compose config file path fix

kavach:
### What
The config path was incorrect and it created an empty directory instead of the config file.

### Why
To run the container correctly.

### Changes
 - Updated the service -app in docker-compose.yml file
 
### Tested
- Local Docker setup

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -15,7 +15,7 @@ services:
         condition: service_healthy
     volumes:
       - ./uploads:/app/uploads:rw
-      - ./config.toml:/app/config.toml:ro
+      - ../config.toml:/app/config.toml:ro
     command: ["server", "-migrate", "-config", "/app/config.toml"]
 
   # PostgreSQL database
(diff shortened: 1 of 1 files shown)


shridarpatil:
Thanks for the PR! Just to clarify, the Docker files are created to run like this:

```bash
# Download compose file and sample config
curl -LO https://raw.githubusercontent.com/shridarpatil/whatomate/main/docker/docker-compose.yml
curl -LO https://raw.githubusercontent.com/shridarpatil/whatomate/main/config.example.toml

# Copy and edit config
cp config.example.toml config.toml

# Run services
docker compose up -d
```

This approach handles all dependencies (PostgreSQL, Redis, etc.) automatically.

ankit-makwana:
@shridarpatil I was trying to follow **Quick Start (Docker)** guide from Docs and i faced issue mentioned above. Then i tried steps according to your comment and it worked.

shridarpatil:
ah I need to update the docs. I readme, it's correct, but I missed updating the docs.

PR #65, opened 2026-01-23
Added railway deployment template to README.md

brunocascio:
I've created a Railway template so people can deploy it easily at railway without the hassle of configuring it :)


shridarpatil:
Can you add this to the documentation instead of the README?

brunocascio:
done @shridarpatil! 

Let me know if I should move it to another section

PR #98, opened 2026-01-28
Fix: ensure notification toasts respect dark mode settings

unnikrishnanNam:
# fix(ui): sync notification toast theme with application color mode

## Description
This PR fixes an issue where notification toasts (using `vue-sonner`) remained in light mode even when the application was switched to dark mode. The `Toaster` component was missing the `theme` prop binding, causing it to ignore the application's current color scheme.

## Changes
- Updated `App.vue` to destructure `colorMode` from the `useColorMode` composable.
- Bound the `colorMode` state to the `theme` prop of the `<Toaster />` component.

## 📸 Screenshots

| Before (Dark Mode with Light Toast) | After (Dark Mode with Dark Toast) |
|:-----------------------------------:|:---------------------------------:|
| Before Fix | After Fix |

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -5,14 +5,14 @@ import { TooltipProvider } from '@/components/ui/tooltip'
 import { useColorMode } from '@/composables/useColorMode'
 
 // Initialize color mode early
-useColorMode()
+const { colorMode } = useColorMode()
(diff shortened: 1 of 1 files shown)


shridarpatil:
Can you add gif/screenshots?

PR #152, opened 2026-02-06
Ability to create multiple organizations Ability to add or assign users to one or more organizations

Deepu97:
This update introduces multi-organization support in the admin panel. Admins can now create and manage one or more organizations, and while creating or editing a user, they can assign that user to any selected organization from the list of organizations created in the system.


shridarpatil:
This is already implemented in https://github.com/shridarpatil/whatomate/pull/145

PR #160, opened 2026-02-10
perf: optimize campaign stats update to eliminate N+1 queries

unnikrishnanNam:
### The Problem
Currently, every campaign status update (message delivered, read, etc.) triggers a database `UPDATE`, followed immediately by a `SELECT` query just to retrieve the `OrganizationID` needed for the WebSocket broadcast.

For a campaign with 100,000 recipients, processing webhooks results in 100,000 completely unnecessary `SELECT` queries during peak load.

### The Solution
I've refactored `incrementCampaignStat` in `internal/handlers/campaigns.go` to use GORM's `Clauses(clause.Returning{})`. This leverages the Postgres `RETURNING` clause to fetch the updated record state (including `OrganizationID` and new counts) in the same atomic transaction as the update.

### Benefits
- **Halves DB load:** Reduces database hits by 50% per webhook event.
- **Atomic:** Ensures the projected WebSocket data matches exactly what was just committed.
- **Scalability:** Significantly reduces database pressure during large campaign blasts.

### Verification
- [x] Code compiles
- [x] Existing handler tests pass

### DIFF
--- a/cmd/whatomate/main.go
+++ b/cmd/whatomate/main.go
@@ -207,6 +207,7 @@ func runServer(args []string) {
 	g.Before(middleware.SecurityHeaders())
 	g.Before(middleware.RequestLogger(lo))
 	g.Before(middleware.Recovery(lo))
+	g.Before(middleware.CSRFProtection())
 
 	// Setup routes
 	setupRoutes(g, app, lo, cfg.Server.BasePath, rdb, cfg)
@@ -432,6 +433,7 @@ func setupRoutes(g *fastglue.Fastglue, app *handlers.App, lo logf.Logger, basePa
 	}
 	g.POST("/api/auth/logout", app.Logout)
 	g.POST("/api/auth/switch-org", app.SwitchOrg)
+	g.GET("/api/auth/ws-token", app.GetWSToken)
 
 	// SSO routes (public)
 	g.GET("/api/auth/sso/providers", app.GetPublicSSOProviders)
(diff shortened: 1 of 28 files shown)


shridarpatil (commented):
Good optimization — using `RETURNING` to fold two queries into one is the right approach for this hot path. One issue to address:

### Bug: broadcast on non-existent campaign

The old code had a natural guard — if the campaign didn't exist, the `SELECT` would fail and the broadcast was skipped (`if err == nil`). With the new code, if the `UPDATE ... RETURNING` matches 0 rows (campaign doesn't exist or was deleted), GORM won't populate the struct, so `campaign.OrganizationID` stays as the zero UUID and the broadcast fires to a phantom org.

Fix by checking `RowsAffected`:

```go
result := a.DB.Model(&campaign).
    Clauses(clause.Returning{}).
    Where("id = ?", campaignUUID).
    Update(column, gorm.Expr(column+" + 1"))
if result.Error != nil {
    a.Log.Error("Failed to increment campaign stat", "error", result.Error, "campaign_id", campaignID, "column", column)
    return
}

// Broadcast stats update via WebSocket
if a.WSHub != nil && result.RowsAffected > 0 {
```

### Nit: redundant WHERE clause

`.Model(&campaign)` with `campaign.ID` already set causes GORM to add `WHERE id = ?` automatically. The explicit `.Where("id = ?", campaignUUID)` is redundant. Not harmful, but you could drop it for clarity.

---

Otherwise LGTM — the `RETURNING` approach is atomic, correct, and the import is clean.

unnikrishnanNam:
Thanks for the review and pointing out that part. I'll push an updated commit shortly.

unnikrishnanNam:
**Changes**
*   Refactored `incrementCampaignStat` to separate the DB operation from the WebSocket broadcast.
*   Replaced the separate `SELECT` query with `Clauses(clause.Returning{})` in the `UPDATE` statement.
*   Added a check for `RowsAffected > 0` before broadcasting. This fixes a bug where we were sending empty WebSocket events for non-existent or deleted campaigns.
*   Removed a redundant `.Where()` clause (GORM handles this via the Model ID).

PR #405, opened 2026-05-17
Implement sending WhatsApp Flows from chat interface

Saifallak:
# Still need tests, but works with manuel tests.


shridarpatil:
Why do we need separate functionality to send WhatsApp flow? This can be handled by a template we can create template with a flow button, and from chat, we can send that template, which is already implemented.

Saifallak:
> Why do we need separate functionality to send WhatsApp flow? This can be handled by a template we can create template with a flow button, and from chat, we can send that template, which is already implemented.

Sending template costs money even in time window thats already open,

But sending flow will count as normal message and costs zero

shridarpatil:
Utility templates delivered within an open customer service window are free as per whatsapp documentation.
Check the docs: https://developers.facebook.com/documentation/business-messaging/whatsapp/pricing

image

Saifallak:
still, sending template requires pre-approval from meta, while flows don't as far as i know.
it's okay if you don't agree with this, it's not critical, just thought it's nice to have.

shridarpatil:
too many options make the UX bad. You can add flow option in canned response if you really dont want to send templates that will solve the problem without adding another option to send flows

Saifallak:
> too many options make the UX bad. You can add flow option in canned response if you really dont want to send templates that will solve the problem without adding another option to send flows

didn't think of that.
--
i checked now, and there is no option to send flow, can you explain more ?.


image


send flow only available in templates.

shridarpatil:
https://github.com/shridarpatil/whatomate/pull/470 fixed here

PR #127, opened 2026-02-03
Feature: Embedded Signup

Saifallak:
Original PR: https://github.com/Saifallak/whatomate/pull/6

# Feature: WhatsApp Embedded Signup Integration

This PR implements the full Embedded Signup flow for WhatsApp Business Platform, allowing users to connect their accounts directly through the application using their Facebook credentials.

## Frontend Changes
- **AccountsView.vue**:
  - Added "Connect with Facebook" button triggered by [launchWhatsAppSignup](file:///Users/saifallahkhaled/PhpstormProjects/whatomate/frontend/src/views/settings/AccountsView.vue#176-252).
  - Implemented dynamic Facebook SDK loading and initialization ([loadFacebookSDK](file:///Users/saifallahkhaled/PhpstormProjects/whatomate/frontend/src/views/settings/AccountsView.vue#157-175)).
  - Added logic to handle the Facebook Login popup, capture the authorization code (`authResponse.code`), and send it to the backend.
  - Implemented [exchangeCodeForToken](file:///Users/saifallahkhaled/PhpstormProjects/whatomate/frontend/src/views/settings/AccountsView.vue#253-291) to handle the backend response.
  - Added UI support for "Pending Registration" state (allowing manual PIN entry later) and success notifications showing the generated 2FA PIN.
  - Added "Test Number" and "Manual Entry" badge support.

## Backend Changes
- **New API Endpoint (`POST /accounts/exchange-token`)**:
  - Handles the OAuth 2.0 flow by exchanging the authorization code for a long-lived Meta access token.
  - **Account Discovery**: Implemented intelligent discovery logic to find the WABA ID and Phone Number ID directly from the access token debug info (checking granular scopes) if they are not provided by the frontend.
  - **Auto-Registration**: The system now attempts to automatically register the phone number with Meta using a random 6-digit PIN immediately upon connection to streamline the user experience.
  - **Webhook Subscription**: Automatically subscribes the app to webhooks for the discovered WABA to ensure message delivery.
  - **Validation**: Validates the phone number status and ensures it belongs to the WABA.

- **New API Endpoint (`POST /accounts/{id}/register`)**:
  - Allows for manual Two-Step Verification (registration) if the auto-registration fails or if the user needs to set a custom PIN.

- **Infrastructure ([pkg/whatsapp/client.go](file:///Users/saifallahkhaled/PhpstormProjects/whatomate/pkg/whatsapp/client.go))**:
[...]

PR #87, opened 2026-01-26
[WIP] Embedded SignUp

Saifallak:
This would the following
- Fixes https://github.com/shridarpatil/whatomate/issues/35
- Fixes https://github.com/shridarpatil/whatomate/issues/20

PR #415, opened 2026-05-21
Feat: Embedded Signup and Coexistence Support Using Meta Flows.

### LINKED ISSUE #20: Support for WhatsApp Coexistence
Requesting support for WhatsApp Coexistence, allowing the same WhatsApp number to be used simultaneously with both the WhatsApp Business App and the WhatsApp Cloud API integration.

### LINKED ISSUE #35: Feature Req: Whatsapp Embedded Sign Up
**Is your feature request related to a problem? Please describe.**
The problem is a complicated sign-up procedure that requires technical intervention involving facebook developer account.

**Describe the solution you'd like**
Integrate Whatsapp Embedded Sign Up so that user's can easily onboard their whatsapp number from a single popup window and few steps.

### LINKED ISSUE #411: Feature Request: WhatsApp Coexistence
Meta rolled out **WhatsApp Coexistence** in May 2025, allowing businesses to use the WhatsApp Business App and Cloud API on the **same number simultaneously**.

This means:
- No need to migrate away from the existing Business App number
- Messages sent from the mobile app are mirrored into the platform inbox via `smb_message_echoes` webhooks
- Contacts synced via `smb_app_state_sync`
- Optional 6-month chat history import

It uses a modified Embedded Signup flow with `featureType: "whatsapp_business_app_onboarding"`.

This feature is already live in Wati, GoHighLevel, Chatwoot, and several other platforms. It's becoming the standard onboarding path for SMBs on the WhatsApp Business App.

Would be great to see this in Whatomate — it would significantly lower the barrier for new users to adopt the platform.

📖 Meta docs: https://developers.facebook.com/docs/whatsapp/embedded-signup/custom-flows/onboarding-business-app-users

Saifallak:
Original PRs
- https://github.com/shridarpatil/whatomate/pull/121 ( Very Old )
- https://github.com/shridarpatil/whatomate/pull/87 ( Very Old )
- https://github.com/shridarpatil/whatomate/pull/127 ( Had Comments to be addressed )
- https://github.com/Saifallak/whatomate/pull/6 ( Just fork to run tests )

Fixes:
- Fixes https://github.com/shridarpatil/whatomate/issues/20
- Fixes https://github.com/shridarpatil/whatomate/issues/411
- Fixes https://github.com/shridarpatil/whatomate/issues/35

-------


spss20:
When can we expect this to be merged?

Saifallak:
> When can we expect this to be merged?

You can use my fork, if you need it immediately, or wait for pr review by the maintainers

chowdareddytg:
> > When can we expect this to be merged?
> 
> You can use my fork, if you need it immediately, or wait for pr review by the maintainers

@Saifallak any documentation or steps available to setup Embedded Signup

Saifallak:
> > > When can we expect this to be merged?
> > 
> > 
> > You can use my fork, if you need it immediately, or wait for pr review by the maintainers
> 
> @Saifallak any documentation or steps available to setup Embedded Signup

Just bring those info from you Facebook developer app
Screenshot_20260523_165419_Chrome


And add them to your config.toml then inside accounts page, you will find connect with facebook btn click it and follow the steps to link phone 📱 number

chowdareddytg:
> > > > When can we expect this to be merged?
> > > 
> > > 
> > > You can use my fork, if you need it immediately, or wait for pr review by the maintainers
> > 
> > 
> > @Saifallak any documentation or steps available to setup Embedded Signup
> 
> Just bring those info from you Facebook developer app Screenshot_20260523_165419_Chrome
> 
> And add them to your config.toml then inside accounts page, you will find connect with facebook btn click it and follow the steps to link phone 📱 number

How/where to get config_id?

Saifallak:
> > > > > When can we expect this to be merged?
> > > > 
> > > > 
> > > > You can use my fork, if you need it immediately, or wait for pr review by the maintainers
> > > 
> > > 
> > > @Saifallak any documentation or steps available to setup Embedded Signup
> > 
> > 
> > Just bring those info from you Facebook developer app Screenshot_20260523_165419_Chrome
> > And add them to your config.toml then inside accounts page, you will find connect with facebook btn click it and follow the steps to link phone 📱 number
> 
> How/where to get config_id?

Screenshot_20260525_202508_Chrome
Screenshot_20260525_202540_Chrome
Screenshot_20260525_202601_Chrome

shridarpatil:
Thanks for picking this up @Saifallak — the test coverage is genuinely solid and this is a long-overdue feature. A few things to address before merging. Grouping by priority.

## Blocking — security & multi-tenancy

### 1. `account.AccessToken` is stored in plaintext
`ExchangeToken` does:
```go
account.AccessToken = accessToken
```
Every other path encrypts it (`CreateAccount` accounts.go:116, `UpdateAccount` accounts.go:231) and reads via `WhatsAppAccount.DecryptSecrets` (models.go:338). A plaintext value will (a) leak via DB access and (b) corrupt later reads when `DecryptFields` runs against bytes that aren't ciphertext. Please wrap with `crypto.Encrypt(accessToken, a.Config.App.EncryptionKey)`. The assertion `assert.Equal(t, "EAABwzLixnjYBO1234567890", account.AccessToken)` in `TestApp_ExchangeToken_Success_AutoRegistration` actually confirms today's plaintext behaviour and will need updating once encryption is in place.

### 2. `account.AppSecret` is never populated → webhook signature verification silently skipped
The verifier loop at `webhook.go:257-260`:
```go
if err != nil || account.AppSecret == "" { continue }
if !verifyWebhookSignature(body, signature, []byte(account.AppSecret)) { ... }
```
short-circuits when `AppSecret` is empty. The PR never assigns `account.AppSecret` in `createOrUpdateAccount`, so every webhook for an embedded-signup account bypasses HMAC verification with no warning. Please populate it (encrypted) from the resolved Meta app secret at account creation.

### 3. New `Pin` field leaks via JSON and is unencrypted at rest
```go
Pin string `gorm:"size:6" json:"pin"`
```
Every other secret on this model uses `json:"-"` (`AccessToken`, `AppSecret`, `KeyHash`, …). The PIN is the 2FA factor for the Meta WABA — any user with `accounts:read` can read it via `GetAccount` / `ListAccounts`. Please:
- Change tag to `json:"-"`
- Encrypt at rest, extend `DecryptSecrets` to cover it
- Keep the one-time disclosure in the `ExchangeToken` response (the legitimate moment)

### 4. New handlers skip `requirePermission` and `audit.LogAudit`
Every other CRUD handler on this resource gates writes with `requirePermission(r, userID, "accounts", "write")` and emits an audit entry (per CLAUDE.md "Conventions"). `ExchangeToken` and `RegisterPhone` do neither — so any org member (including viewers) can connect a WhatsApp number, and the e2e suite's `verifyAuditLogged()` will fail. Mirror the pattern in `CreateAccount` (accounts.go:88, 164).

### 5. Multi-tenancy: Meta App credentials must be per-org, not global

Whatomate is a multi-tenant platform; locking every tenant on an instance to a single Meta App via `Config.WhatsApp.AppID` / `AppSecret` / `ConfigID` contradicts that. Each org should be able to bring its own Meta App credentials. The codebase already has the right primitive — `Organization.Settings` (JSONB) plus the per-org settings pattern in `internal/handlers/organization.go` (calling settings already use it as "per-org override → global fallback").

Required shape:

1. Add `MetaAppID`, `MetaConfigID` to `OrganizationSettings` (`organization.go:41`). Store `meta_app_secret_encrypted` in the JSONB blob — **write-only**, surface only `has_meta_app_secret: bool` on read, mirroring `AccountResponse.HasAppSecret` (`accounts.go:51`, `accounts.go:476`).
2. Add a `resolveMetaAppCreds(orgID) (appID, appSecret, configID, error)` helper that prefers per-org settings and falls back to `Config.WhatsApp.*` for a SaaS-mode default.
3. Use it in `GetEmbeddedSignupConfig` (scope by JWT orgID), `ExchangeToken`, and `discoverWABAAndPhone` — replacing the three current direct `a.Config.WhatsApp.*` reads.
4. In `createOrUpdateAccount`, only fill `account.AppID` / `APIVersion` if empty — currently the PR overwrites them unconditionally, which silently wipes per-account values set via `UpdateAccount` and is read by `pkg/whatsapp/client.go:426` (`ResumableUpload` for template media uploads).
5. Surface the three fields in the Accounts/Settings UI, gated on `accounts:write`. The PR already adds matching i18n keys for `encryptionConfig` — same pattern.
6. Make `fetchWhatsAppConfig` (`AccountsView.vue`) re-run when `organizationsStore.selectedOrgId` changes — currently fires only in `onMounted`, so the SDK button stays bound to the first-loaded org's app after switching tenants.

---

## Code reuse

### 6. `processMessageEcho` should reuse the content-extraction step from `processIncomingMessageFull` — but not call it directly

The new function re-implements the same text/button/interactive/image/document/video/audio/sticker/location/contacts switch already in `chatbot_processor.go:130-310`, plus duplicate-detection and contact resolution. Two truth sources will silently drift — any new media type added to one path bypasses the other.

**Important: don't just call `processIncomingMessageFull` from the echo handler.** The second half of that function would be actively wrong for an echo:

- It saves the message with `Direction: incoming, Status: received` — must be `outgoing, sent` for an echo.
- It triggers chatbot processing — `getChatbotSettingsCached` → `getOrCreateSession` → `runChatGraph` — which would walk the chatbot graph against the agent's own outbound message. At best wasted work, at worst an automation loop where the bot responds to its own echoes.
- It runs business-hours / out-of-hours auto-reply, keyword-matching, and agent-transfer creation — all keyed on customer intent, none of which apply to an outbound message.
- It dispatches `WebhookEventMessageIncoming`, picks the customer as contact (`msg.From`), and marks the conversation unread — all reversed for an echo (`msg.To`, marked read).

The right shape is to extract just the content-derivation step:

```go
type ExtractedMessage struct {
    Text             string
    Type             string         // may differ from msg.Type, e.g. "button_reply"
    Media            *MediaInfo
    ButtonID         string         // used by chatbot routing only
    FlowResponseData map[string]any // used by chatbot routing only
}

// extractMessageContent walks an IncomingTextMessage and returns the derived
// fields. Used by both the inbound and echo paths.
func (a *App) extractMessageContent(msg IncomingTextMessage, account *models.WhatsAppAccount) ExtractedMessage
```

After this:
- `processIncomingMessageFull` calls `extractMessageContent`, then continues with `saveIncomingMessage` + chatbot machinery as before.
- `processMessageEcho` calls `extractMessageContent`, then builds a message with `Direction: outgoing, Status: sent`, broadcasts, dispatches `message.sent`, and **stops** — no chatbot, no transfer, no session.

`processMessageEcho` shrinks from ~230 lines to ~30–40, and the media switch lives in exactly one place.

### 7. Echo dispatches `WebhookEventMessageOutgoing`, but subscribers can't opt into it
The user-facing event list at `webhooks.go:113` (`AvailableWebhookEvents`) only exposes `WebhookEventMessageSent`. `WebhookEventMessageOutgoing` is referenced only in `goroutines_test.go` — effectively an undocumented event with no UI subscription path. Either dispatch `WebhookEventMessageSent` (with an optional `Source` discriminator if you want to distinguish API-sent from mobile-app echoes) or add `message.outgoing` to `AvailableWebhookEvents`.

### 8. `processContactSync` re-implements `contactutil`'s normalization inline
The "remove" branch does its own `+`-strip and dual-query, which is exactly what `contactutil.GetOrCreateContact` already does (`contactutil.go:20-25`). Same behaviour today, but a future change to normalization lands in one place but not the other. Worth factoring a small `contactutil.FindContact(...)` helper.

### 9. Hand-built map in `ExchangeToken` response
Reimplements `accountToResponse` plus a couple of extras. Could be:
```go
out := map[string]any{"account": accountToResponse(*account)}
if account.Status == "active" && account.Pin != "" { out["pin"] = account.Pin }
if regErr != nil { out["warning"] = "Registration failed: " + regErr.Error() }
return r.SendEnvelope(out)
```

---

## Smaller items

- **`discoverWABAAndPhone`** silently picks `phonesResp.Data[0]` when a WABA exposes multiple numbers. Log a warning when `len > 1`, or return 409 with the list so the SPA can disambiguate.
- **`RegisterPhone`** re-fetches phone info from Meta on every call to detect SMB. Persist `IsSMB` on the account during `ExchangeToken` to avoid the round trip (and the failure mode when Meta is temporarily unreachable).
- **Naming**: handler is `RegisterPhone`, client method is `RegisterPhoneNumber`. Existing convention is the long form (`GetPhoneNumberInfo`, `GetWABAPhoneNumbers`). Pick one.
- **`main.go`** route registration has 2-space indentation in a tab-indented file plus a double blank line above. Won't pass `gofmt`.
- **`client_config_test.go`** constructs `&handlers.App{Config: ...}` directly instead of using `newTestApp(t)`. Will nil-pointer the moment this handler grows a DB call.
- **Test gaps**: no assertion that `account.AccessToken` ends up encrypted (would have caught item 1); no assertion that audit logging fires.
- **Goroutine waits in webhook tests** use `time.Sleep(100 * time.Millisecond)` three times — flaky on loaded CI. `App` already exposes `WaitForBackgroundTasks()` used 10+ times in `goroutines_test.go`.
- **Missing trailing newline** in `frontend/e2e/pages/AccountsPage.ts`, `frontend/src/views/settings/AccountsView.vue`, `internal/models/models.go`.
- **Frontend `(window as any).FB`** — TS-strict project; a minimal `declare global` interface for `window.FB` would be cleaner.
- **Hardcoded brand colors** (`#1877F2`, `#0C5DC7`, `#166FE5`, `#0A4DAD`) in the template — consider promoting to Tailwind theme tokens.
- **Inline i18n fallback** `$t('accounts.connectFacebook', 'Connect with Facebook')` duplicates the dictionary entry the same PR adds. Drop the fallback.

---

## Suggested split

Given the size, possibly easier to merge in slices:
- **PR A**: blockers 1–4 (security + auth/audit). Small, schema-free.
- **PR B**: per-tenant Meta App via `Organization.Settings` (blocker 5).
- **PR C**: the `extractMessageContent` refactor and smaller items.

Happy to review the splits as separate PRs.

Saifallak:
@shridarpatil you mean should i make prs targeting this pr branch? And merge one by one? Then merge this pr?

shridarpatil:
Thanks for consolidating this — the structure is solid, multitenancy holds up (per-org credential resolution outbound, per-account app-secret verification inbound), and the test coverage is genuinely good. A few things to address before merge:

**Lint/formatting (CI blockers — `make lint` will fail)**
- `internal/models/models.go` lost its trailing newline (the `AuditLog.TableName` close).
- New `WhatsAppAccount` fields `IsSMB`/`Pin` aren't gofmt-aligned with the surrounding tag block.
- Stray double blank lines in `webhook.go` (after the `extractMessageContent` assignment in `processMessageEcho`), `contactutil.go` (trailing), and the end of `embedded_signup_test.go`.
- `AccountsView.vue` and `AccountsPage.ts` lost trailing newlines.

Please run `make fmt`, `make lint`, and `npm run lint`.

**Correctness**
1. **`createOrUpdateAccount` always sets `Status = "pending_registration"`**, even on the existing-account branch. Re-running embedded signup for an already-active number resets it to pending; if the subsequent re-registration then fails, the account is left downgraded. Preserve the prior status on the existing-account path unless re-registration succeeds.
2. **No panic recovery in `go a.processMessageEcho` / `go a.processContactSync`** — these run on untrusted webhook input, and a panic in a goroutine isn't caught by the Recovery middleware. A `defer recover()` would be safer.

**Inconsistent default API version**
Manual `CreateAccount` hardcodes `"v21.0"` while the embedded-signup path uses `Config.WhatsApp.APIVersion` (for FB SDK init, token exchange, discovery, and the stamped account default). So within the same org, two accounts can land on different Graph API versions depending on how they were created — e.g. if config is `v24.0`, manual accounts silently get v21.0. Both creation paths should resolve the default version the same way. (Per-org API version isn't needed — the Graph version is a deployment concern and each account stores its own value; this is purely about the inconsistent default.)

**Maintainability — duplicated logic**
- **`createOrUpdateAccount` vs `CreateAccount`:** a separate helper is justified (embedded signup needs upsert semantics, discovered inputs, and deferred encryption so the plaintext token survives for the register/subscribe calls — none of which the HTTP handler `CreateAccount` can provide). But the two paths now reimplement verify-token generation, the default-version fallback, and the encrypt-secrets block independently (hence the version drift above). Please extract the shared primitives into common helpers both call, so the manual and embedded paths can't diverge.
- **Triplicated message struct:** `MessageEchoes` is a third hand-rolled copy of the message field shape already defined as `IncomingTextMessage` and as `WebhookPayload.Messages`. `processMessageEcho` then bridges it via a `json.Marshal`→`json.Unmarshal` round-trip, which silently drops any mismatched field. Reuse the `IncomingTextMessage` named type for `MessageEchoes` (and ideally `Messages`) to drop both the extra struct and the round-trip. (The `extractMessageContent` extraction itself is a nice de-dup — keep it.)

**Behavior change to confirm**
`ListAccounts`/`GetAccount` now require `accounts:read`/`write` where they were previously auth-only. Good improvement, but please confirm default roles grant `accounts:read` so existing non-admin members don't suddenly get 403s.

**Scope creep**
The `en.json`/`ar.json` edits are unrelated to embedded signup — notably `common.required` changed from `"Required"` to `"{field} is required"` (could break any standalone usage) and some Arabic IVR keys were removed. Can these be split into a separate PR?

**Test nit**
The new `smb_message_echoes` / `smb_app_state_sync` tests use fixed `time.Sleep(100ms)` to wait on goroutines — `assert.Eventually` (already used elsewhere in this file) would be less flaky under CI load.

The per-org Meta App support, encryption-at-rest handling (token/secret/PIN), and the `extractMessageContent` refactor all look great.

Saifallak:
Thanks for your kind review, i will work on those notes soon (maybe tonight or tomorrow)

Saifallak:
@shridarpatil done, each commit addresses one issue,

PR #416, opened 2026-05-21
Fix Campaign Read Counts and Propagate Recipient Failure Reasons

### LINKED ISSUE #421: Campaign detail page does not show total Read.
In the campaign list page, it shows the total number of read but in the campaign detail page the Read shows zero.

Saifallak:
## Description
This PR resolves two issues related to bulk campaign tracking:
1. **Always-zero Read Counts**: Correctly populates and returns the campaign's `ReadCount` field in the API handlers for campaign creation, retrieval, and updating.
2. **Missing Failure Reasons**: Propagates the failure error message from WhatsApp status update webhooks down to the individual campaign recipient record (`BulkMessageRecipient`) and displays it in the frontend UI under the failed recipient's status badge.

---

### Backend
1. **API Response Mappings**:
   - File: [`internal/handlers/campaigns.go`](/whatomate/internal/handlers/campaigns.go)
   - Populated the missing `ReadCount: campaign.ReadCount` fields mapping to `CampaignResponse` in `CreateCampaign`, `GetCampaign`, and `UpdateCampaign`.

2. **Webhook Error Propagation**:
   - File: [`internal/handlers/webhook.go`](/whatomate/internal/handlers/webhook.go)
   - Modified `updateMessageStatus` to copy the `error_message` from message status updates to the corresponding `BulkMessageRecipient` record when a message fails.

### Frontend
1. **UI Display of Recipient Errors**:
   - File: [`frontend/src/views/settings/CampaignDetailView.vue`](/whatomate/frontend/src/views/settings/CampaignDetailView.vue)
   - Updated the recipient list table to render `recipient.error_message` below the "failed" badge when message delivery fails.
   - Truncated the message at 180px for a clean look, displaying the full message in a native browser tooltip on hover.

---

### Unit Tests Added
1. **Campaign Response Verification**:
   - File: [`internal/handlers/campaigns_test.go`](/whatomate/internal/handlers/campaigns_test.go)
   - Updated `TestApp_CreateCampaign_Success`, `TestApp_GetCampaign_Success`, and `TestApp_UpdateCampaign_Success` to assert the correctness of the returned `ReadCount` mapping.
2. **Recipient Error Message Assertion**:
   - File: [`internal/handlers/webhook_test.go`](/whatomate/internal/handlers/webhook_test.go)
   - Updated `TestUpdateMessageStatus_FailedUpdatesMessage` to verify that `recipient.ErrorMessage` is successfully populated on status failure.

### Test Status
All tests passed successfully:
```bash
go test ./...
# Output:
# ok  	github.com/shridarpatil/whatomate/internal/handlers	0.755s
```

Fixes #421


internal/handlers/campaigns_test.go
```diff
@@ -151,6 +151,7 @@ func TestApp_CreateCampaign_Success(t *testing.T) {
 	assert.Equal(t, "Test Campaign", resp.Data.Name)
 	assert.Equal(t, models.CampaignStatusDraft, resp.Data.Status)
 	assert.Equal(t, template.ID, resp.Data.TemplateID)
+	assert.Equal(t, 0, resp.Data.ReadCount)
```
shridarpatil: passes with or without the fix, so it doesn't guard the regression. Not worth holding the PR; you could drop that one line or update the test with read data

shridarpatil: Actually, you can drop this as there is already a test on line 291

Saifallak: > Actually, you can drop this as there is already a test on line 291

removed that line, and yes there is two tests have numbers

PR #417, opened 2026-05-21
 WhatsApp Connection Health & Template Quality Metrics

Saifallak:
This Pull Request introduces connection health diagnostic details on WhatsApp Business Account connection validation (Test Connection) and template quality rating metrics across the application settings, fully localized in Arabic and English.

---

### 🗄️ Backend
- **Models**: Modified `internal/models/models.go` to add `QualityRating` to the `Template` struct.
- **Client**: Modified `pkg/whatsapp/types.go` to add `QualityRating` field to `MetaTemplate`.
- **Handlers**:
  - Modified `internal/handlers/templates.go` to fetch, map, and persist the quality rating from Meta during sync.
  - Modified `internal/handlers/accounts.go` / validation logic to return full connection details (Verified Name, Quality Rating, Messaging Limit Tier, Verification Status) inside the test connection response payload.
- **Tests**: Updated `accounts_validation_test.go` and template tests to assert mocked connection fields.

### 🎨 Frontend
- **Account Details Settings**:
  - Updated `AccountDetailView.vue` to show a new panel under connection validation result containing Verified Name, Quality Rating (badge), Messaging Limit, and Verification Status (badge).
  - Integrated full translation support for connection metrics so that raw API strings (e.g. `GREEN`, `TIER_10K`) are represented in user-friendly localized text.
- **Templates List**:
  - Updated `TemplatesView.vue` data table to include a new **Quality Rating** column.
  - Styled with CSS variables mapping to premium light/dark HSL background/text colors depending on status (Green/High, Yellow/Medium, Red/Low).
- **Template Detail & Preview**:
  - Updated `TemplateDetailView.vue` to render the quality rating badge next to the template status in the main card header.
  - Added the Quality Rating row to the template preview dialog.

### 🌐 Translations (Arabic & English)
Locales are updated in `ar.json` and `en.json` to map quality levels and connection statuses:

| Key | Arabic (ar) | English (en) |
| :--- | :--- | :--- |
| `accounts.qualityGreen` | أخضر (عالي) | Green (High) |
| `accounts.qualityYellow` | أصفر (متوسط) | Yellow (Medium) |
| `accounts.qualityRed` | أحمر (منخفض) | Red (Low) |
| `accounts.qualityUnknown` | غير معروف | Unknown |
| `accounts.statusVerified` | تم التحقق | Verified |
| `accounts.statusNotVerified` | لم يتم التحقق | Not Verified |
| `accounts.statusExpired` | منتهي | Expired |

---

### 1. Go Backend Tests
All tests compile and pass successfully:
```bash
go test -count=1 ./pkg/whatsapp/... ./internal/handlers/...
# Output:
# ok  	github.com/shridarpatil/whatomate/pkg/whatsapp	0.590s
# ok  	github.com/shridarpatil/whatomate/internal/handlers	0.458s
```

### 2. Frontend Build
- The typescript compilation was validated via `npm run typecheck` and passes with zero compilation errors in the modified files.

### Screenshots:

after clicking on test in account detail view.

( account got blocked so verification expired )
image

(working account)
image

(meta testing number, it's doesn't get verified anyway)
image

(smb account)
image

inside templates

image


frontend/src/i18n/locales/en.json
```diff
@@ -1259,6 +1261,19 @@
...
   "accounts": {
+    "metaConnectionDetails": "Meta Connection Details",
+    "verifiedName": "Verified Name",
+    "qualityRating": "Quality Rating",
+    "messagingLimitTier": "Messaging Limit Tier",
+    "accountMode": "Account Mode",
+    "codeVerificationStatus": "Verification Status",
+    "qualityGreen": "Green (High)",
+    "qualityYellow": "Yellow (Medium)",
+    "qualityRed": "Red (Low)",
```
shridarpatil: Why show color name? Just add high medium and low. And based on the quality we can show badge color.

frontend/src/views/settings/AccountDetailView.vue
```diff
@@ -133,6 +134,75 @@ const webhookUrl = window.location.origin + basePath + '/api/webhook'
 // Track form changes
 watch(form, () => { hasChanges.value = true }, { deep: true })
 
+function getQualityBadgeClass(rating: string) {
+  switch (rating.toUpperCase()) {
+    case 'GREEN':
+    case 'HIGH':
```
shridarpatil: why switch for both green and high? Same for other cases.

frontend/src/views/settings/AccountDetailView.vue
```diff
@@ -133,6 +134,75 @@ const webhookUrl = window.location.origin + basePath + '/api/webhook'
...
+    case 'YELLOW':
+    case 'MEDIUM':
+      return t('accounts.qualityYellow')
+    case 'RED':
+    case 'LOW':
+      return t('accounts.qualityRed')
+    default:
+      return rating
+  }
+}
```
shridarpatil: Why do we need this map? We can directly show rating as HIGH, MEDIUM and LOW?

frontend/src/views/settings/TemplateDetailView.vue
```diff
@@ -358,6 +359,40 @@ const statusVariant = computed(() => {
...
+    case 'MEDIUM':
+      return t('accounts.qualityYellow')
+    case 'RED':
+    case 'LOW':
+      return t('accounts.qualityRed')
+    default:
+      return rating
+  }
+}
+
```
shridarpatil: Same with this, we don't need to have a map.

frontend/src/views/settings/TemplatesView.vue
```diff
@@ -289,6 +291,40 @@ function getHeaderIcon(type: string) {
       return MessageSquare
   }
 }
+
+function getQualityBadgeClass(rating: string) {
```
shridarpatil: Duplicate method. Why don't we reuse the method used in the template details view

frontend/src/views/settings/TemplatesView.vue
```diff
@@ -289,6 +291,40 @@ function getHeaderIcon(type: string) {
...
+      return 'bg-yellow-950 text-yellow-400 border border-yellow-800/40 light:bg-yellow-100 light:text-yellow-800'
+    case 'RED':
+    case 'LOW':
+      return 'bg-red-950 text-red-400 border border-red-800/40 light:bg-red-100 light:text-red-800'
+    default:
+      return 'bg-gray-800 text-gray-400 light:bg-gray-100 light:text-gray-600'
+  }
+}
+
+function getQualityRatingLabel(rating: string) {
```
shridarpatil: same duplicate method.

Saifallak:
i will address your comments tomorrow, and mark it as ready for review.

Saifallak:
I have reviewed and confirmed that the refactoring of the WhatsApp quality ratings has been successfully completed and verified:

### 1. Translation Cleanup (`en.json` & `ar.json`)
* **Request:** *"Why show color name? Just add high medium and low. And based on the quality we can show badge color."*
* **Resolution:** The locale files [en.json](/whatomate/frontend/src/i18n/locales/en.json) and [ar.json](/whatomate/frontend/src/i18n/locales/ar.json) have been updated to display only the rating names (e.g. `"High"`, `"Medium"`, `"Low"` / `"عالي"`, `"متوسط"`, `"منخفض"`). Color styling is now handled entirely through the badge classes.

---

### 2. Dual Switch Cases (`GREEN` vs `HIGH`)
* **Request:** *"why switch for both green and high? Same for other cases."*
* **Resolution:** The Meta WhatsApp Business APIs return different raw value representations depending on the resource:
  * **WhatsApp Account/Phone Number Quality** uses colors: `GREEN`, `YELLOW`, and `RED`.
  * **Message Template Quality** uses rating level names: `HIGH`, `MEDIUM`, and `LOW`.
  * The switch handles both sets of values so that both phone quality and template quality can share the same underlying logic.

---

### 3. Displaying Normalized Quality Rating Labels
* **Request:** *"Why do we need this map? We can directly show rating as HIGH, MEDIUM and LOW?"*
* **Resolution:** The mapping is necessary for two reasons:
  1. **Normalization:** It translates phone number colors (e.g. `GREEN`) into the corresponding rating levels (e.g. `High`) to provide a consistent look and feel across different parts of the settings dashboard.
  2. **Localization (i18n):** It enables translated text (e.g., rendering `"عالي"` instead of the English API constant `"HIGH"` when using the app in Arabic).

---

### 4. Deduplicating Local View Helper Methods
* **Request:** *"Duplicate method. Why don't we reuse the method used in the template details view"*
* **Resolution:** The duplicate functions `getQualityBadgeClass` and `getQualityRatingLabel` have been extracted into the shared utility file [utils.ts](/whatomate/frontend/src/lib/utils.ts) and imported in:
  * [AccountDetailView.vue](/whatomate/frontend/src/views/settings/AccountDetailView.vue)
  * [TemplateDetailView.vue](/whatomate/frontend/src/views/settings/TemplateDetailView.vue)
  * [TemplatesView.vue](/whatomate/frontend/src/views/settings/TemplatesView.vue)

shridarpatil:
Nice feature — really useful surface for ops to spot a degraded WABA at a glance. A few items worth addressing before merge:

  🔴 Bugs

  1. getQualityRatingLabel doesn't localize "UNKNOWN" (frontend/src/lib/utils.ts)

  New templates get QualityRating = "UNKNOWN" (set in CreateTemplate and as the sync fallback). The helper's if (!rating) guard skips truthy "UNKNOWN", so the switch falls through and returns the raw string — users see English UNKNOWN even in Arabic, instead of
  the accounts.qualityUnknown key you added. Fix:

  case '':
  case 'UNKNOWN':
    return t('accounts.qualityUnknown')

  2. MetaTemplate.QualityRating is dead in production (pkg/whatsapp/template.go)

  The new fields= whitelist in FetchTemplates requests quality_score but not quality_rating. Meta filters response fields strictly, so metaTemplate.QualityRating will always be empty at runtime — the first line of the conditional in SyncTemplates is a no-op. The
  test passes only because the mock returns the field unconditionally, regardless of the fields= query. Either drop the QualityRating JSON field on MetaTemplate (and the conditional), or add quality_rating to the fields= list.

  3. WABA fallback request has no context/timeout (internal/handlers/accounts.go)

  http.NewRequest(...) (no WithContext) means a slow Meta response silently doubles the "Test Connection" latency. Please propagate r.RequestCtx via http.NewRequestWithContext so fastglue's deadline applies.

  🟡 Nits

  - getQualityBadgeClass is shared in lib/utils.ts but getVerificationBadgeClass and formatLimitTier stay local — pick one convention.
  - Each component re-wraps the shared helper as sharedGetQualityRatingLabel → local getQualityRatingLabel. Just use the shared one with t as the second arg.
  - Stray unrelated field: business_calling_enabled?: boolean added to WhatsAppAccount interface in AccountDetailView.vue but never referenced in this PR.
  - Use http.StatusOK instead of 200 in the WABA fallback check.
  - default:'UNKNOWN' on the GORM tag won't backfill existing rows — they'll have "" until next sync. Worth a line in the PR description about the upgrade story.
  - Branch name typo: feat/quailty-rating → feat/quality-rating.

  🧪 Coverage gap

  No frontend e2e for the new details panel/badges/column. Per frontend e2e, the convention is UI-driven tests with API side-channels — a small Playwright spec driving "Test Connection" and asserting the four fields render would protect the conditional helpers from future regressions. Backend coverage (TestApp_TestAccountConnection_FallbackToWABA, the sync mapping test) is solid.

  🟢 What's good

  Localization is thorough (both AR/EN for every new key — once #1 is fixed, the unused key activates), audit logging is automatic via the existing struct-diff, badge color scheme maps cleanly to Meta's GREEN/YELLOW/RED, and the nested quality_score.score shape
  is covered in the sync test.

  Happy to re-review once the three items are addressed.

Saifallak:
@shridarpatil i don't understand why it fails on CICD but works locally, after solving

> WABA fallback request has no context/timeout (internal/handlers/accounts.go)
> http.NewRequest(...) (no WithContext) means a slow Meta response silently doubles the "Test Connection" latency. Please propagate r.RequestCtx via http.NewRequestWithContext so fastglue's deadline applies.

Saifallak:
Great work @shridarpatil

Saifallak:
But i didn't understand why it was falling, can you tell me?

shridarpatil:
There was a panic error.

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xe02fe0]

github.com/valyala/fasthttp.(*RequestCtx).Done(0xc0001b8008)
    fasthttp@v1.58.0/server.go:2747 +0x40
context.(*cancelCtx).propagateCancel(...)
context.WithDeadlineCause(...)
net/http.setRequestCancel(...)
net/http.(*Client).do(...)
github.com/shridarpatil/whatomate/internal/handlers.(*App).TestAccountConnection
    internal/handlers/accounts.go:369

The test binary crashed mid-suite, killing ~40 other tests as collateral (they show FAIL (-1.00s) because they never got to run).

The line that caused it

The PR changed this:
req, err := http.NewRequest(http.MethodGet, url, nil)
to this: 
req, err := http.NewRequestWithContext(r.RequestCtx, http.MethodGet, url, nil)

Looks harmless. r.RequestCtx is *fasthttp.RequestCtx, which does satisfy the context.Context interface (it has Deadline(), Done(), Err(), Value() methods). So Go compiles it without complaint.

Why it panics

Tracing the stack from the bottom up:

1. a.HTTPClient.Do(req) → stdlib net/http internals.
2. net/http.Client.Do calls setRequestCancel, which always calls context.WithDeadline(req.Context(), …) to attach the client's request timeout.
3. context.WithDeadline calls propagateCancel(parent, child) to wire child cancellation when the parent is cancelled.
4. propagateCancel calls parent.Done() — the parent here is our *fasthttp.RequestCtx.
5. (*fasthttp.RequestCtx).Done() at fasthttp's server.go:2747 reads internal state — specifically the s field (a back-reference to the fasthttp Server that owns the request). It uses s to derive the done-channel.
6. In a real production request, fasthttp's accept loop populates s when it creates the RequestCtx. Works fine.
7. In a test, accounts_validation_test.go builds the request by hand:
r := &fastglue.Request{RequestCtx: &fasthttp.RequestCtx{}}
7. That bare &fasthttp.RequestCtx{} has s == nil. The first dereference inside Done() → SIGSEGV.

The underlying principle

*fasthttp.RequestCtx is a context.Context only by accident of interface satisfaction. It is not safe to hand to anything outside of fasthttp's own server machinery — the methods (Done, Deadline, Value) assume invariants that only hold inside the fasthttp
accept-loop. stdlib net/http doesn't know that and uses it like any other context.

This is a known footgun in mixed fasthttp + stdlib code. Other production handlers in whatomate avoid it — they use plain http.NewRequest(...) (which uses context.Background() internally) and rely on the shared a.HTTPClient for timeouts.

Saifallak:
@shridarpatil thanks for the explanation

PR #428, opened 2026-05-25
fix: use double underscore as env var hierarchy separator

spss20:
The env var override mapping replaced every `_` with `.`, which broke any config key containing an underscore (ssl_mode, read_timeout, access_expiry_mins, etc.). Switch to `__` as the hierarchy separator so single underscores in key names are preserved.

Also update docker-compose.yml to support running without config.toml (env-vars-only mode), which is needed for container platforms like Dokploy or Railway where file mounts aren't practical.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -152,9 +152,10 @@ func Load(configPath string) (*Config, error) {
 	}
 
 	// Load from environment variables (WHATOMATE_ prefix)
-	// e.g., WHATOMATE_DATABASE_HOST -> database.host
+	// Uses __ (double underscore) as hierarchy separator so single underscores in
+	// key names are preserved. e.g., WHATOMATE_DATABASE__SSL_MODE -> database.ssl_mode
 	if err := k.Load(env.Provider("WHATOMATE_", ".", func(s string) string {
-		return strings.ReplaceAll(strings.ToLower(strings.TrimPrefix(s, "WHATOMATE_")), "_", ".")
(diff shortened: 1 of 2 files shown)


shridarpatil:
Thanks for digging in here — and you're right that the original `_ → .` env-var transform was silently broken for any key with an underscore in it (`database.ssl_mode`, `rate_limit.enabled`, `storage.s3_bucket`, etc. — the docs even listed examples that never actually worked).

That said, I'd rather not expand the env-var override surface area further. `config.toml` is the canonical configuration interface for whatomate; the env-var override was meant as a small escape hatch for a handful of secrets, not a full alternate config path. A few specific concerns with this PR in that direction:

- It's a silent breaking change for the one combination that *did* work before — `WHATOMATE_DATABASE_HOST=...` (single underscore, single-word key) would now land under flat key `database_host` and be ignored, with no error.
- The `docker-compose.yml` change drops the `config.toml` mount and the `-config` flag, making env-vars-only the default for the docker workflow.
- The ~90 new `WHATOMATE_*__*` entries in `.env.example` become part of the documented surface area we then have to keep in sync with the struct.

For Dokploy/Railway-style platforms, the supported path is to mount `config.toml` via the platform's secrets/files mechanism. I'll separately fix the docs to stop advertising the broken env-var table.

Leaving this open for now in case you want to scope it down (e.g., just the `internal/config/config.go` + test fix, without the docker-compose / .env.example expansion), but not planning to merge as-is.

shridarpatil:
fixed here https://github.com/shridarpatil/whatomate/pull/478

PR #442, opened 2026-05-29
Deploy/do app

mayankmishra0403:


mayankmishra0403:
addedd ccode

shridarpatil:
Thanks for the contribution @mayankmishra0403, but I'm going to close this.

The bulk of the diff is a rebrand from "Whatomate" to "Ritam Bharat" — 134 lines across i18n locales, README, binary name, Docker image tags, runtime log fields, webhook user-agent, test assertions, and frontend assets. That's a downstream fork concern and can't land in the upstream repo.

The deployment artifacts (`.env` with committed credentials, `app.yaml` pinned to `mayankmishra0403/whatomate`, `bridge/` k8s manifests with `/Users/mayankmishra7296/Desktop/...` hostPaths, `deploy-production.yml` SSH-deploying to your `PROD_*` secrets) only work for your environment and shouldn't live here either.

On the two real code changes:

- **8s polling in `ChatView.vue`**: not needed. Realtime is already covered by the WebSocket connection (`src/services/websocket.ts` + `WSHub`) — polling on top of that is redundant load on the messages endpoint with no benefit.
- **`normalizeOutgoingMessage`** in `contacts.ts`: if outgoing messages are showing up without a `status` in the UI, that's a server-side wire-format question (the API returns `MessageStatusPending` today). Worth a small issue describing what you observed, but not a frontend override.

Closing — please keep the rebrand and deployment infra in your fork. Thanks!

PR #460, opened 2026-06-12
feat: implement Set Variable node and trigger_message session variable

Saifallak:
## Summary
This Pull Request introduces the frontend flow builder interface for the `set_variable` node, registers it in the node palette, implements seeding of trigger and last message variables on session startup, and adds comprehensive unit tests.

### Importance & Value
Previously, chatbot sessions were stateless and could not store ad-hoc information or track user trigger keywords across node execution steps. This update provides:
1. **State Management**: Authors can now save custom keys and values at any point in the flow using the "Set Variable" node.
2. **Context Seeding**: Downstream nodes can reference user inputs or metadata (e.g., in templates or webhooks) dynamically.
3. **Trigger Tracking**: The system keeps a record of the keyword that initiated the flow, allowing context-aware API calls or logging.

---

#### 1. Custom Variable Assignment (Set Variable Node)
Assign multiple values to session variables using templates or constant values:
- **Variable Name**: `greeting` | **Value**: `Hello {{contact_name}}!`
- **Variable Name**: `tier` | **Value**: `VIP`

#### 2. Accessing Seeded Variables (Template Interpolation)
Use variables anywhere in outgoing messages or API headers:
- `Your account type is {{tier}}. {{greeting}}` resolves to -> `Your account type is VIP. Hello Alice!`

#### 3. Trigger and Last Message Tracking
- **Trigger Message**: If a user initiates a session by sending `"PromoCode50"`, `session.SessionData["trigger_message"]` stores `"PromoCode50"` permanently for the duration of that session.
- **Last Message**: If the flow asks for their email and they reply `"user@example.com"`, `session.SessionData["last_message"]` stores `"user@example.com"`.

---

### Frontend
- **New Component**: `ChatbotSetVariableNode.vue` renders the "Set Variable" node in the chatbot canvas.
- **Node Properties**: Added variable assignments editor to `ChatNodeProperties.vue` to configure key-value pairs (supports template substitution).
- **Flow Builder View**: Registered the `set_variable` node and added it to the sidebar palette.

### Backend
- **Runner**: Modified `runChatGraph` in `chatbot_graph_runner.go` to store the initial message that triggered the flow in `trigger_message` and the latest message in `last_message` inside session variables.

### Tests
- **Unit Tests**: Added unit tests in `chatbot_graph_runner_test.go` to verify:
  - Storing of `trigger_message` and `last_message` upon entry (verified using a generic trigger keyword `"Hello Chatbot"`).


shridarpatil:
We should not have a separate node to set a variable.  In the previous step-based builder, we had a store_as, which used to store the response in a session variable, which we could use to format the message. I am working on it to get that back in the node-based flow.

shridarpatil:
https://github.com/shridarpatil/whatomate/pull/463

Saifallak:
i know, but sometimes we need node to set varriable, i will show you a use case.

first we may fetch some categories from api

image


then map it to varriables

image


now adding a button so the user can select what he wishes

image


image

but in store as, only the title can be stored, but in many cases we want to store another thing, like the id from 2nd picture

so we can use it later for another api call or reservation or whatever,


image


--

image


so in many cases we need dedicated set varriable node, so we can use it for any custom logic.
and i agree in many other cases "store_as" is enough

Saifallak:
also the update trigger_message and last_message, is most needed to know which word triggered the flow and use it for query or anything, in my case, i need to know which promocode started the conversation.

shridarpatil:
In store, can we give the option to store both title and ID? That will solve the problem, right?

Saifallak:
hmmm, i think yes but,

> In a WhatsApp chatbot flow builder, if I use the same dynamic/integer values as Button IDs across different question nodes within the same flow, will it cause any conflicts?

> For example, in the first node (e.g., "Select Hotel"), the button IDs will represent hotel_ids (which will likely be 1, 2, etc.). In a subsequent node (e.g., "Select Room"), the button IDs will represent room_ids (which will also be 1, 2, etc.).

> Will the chatbot engine get confused or route the user incorrectly because two different button nodes in the same flow use the exact same button IDs (1 or 2).

i'm not at my desk right now, so idk the answer, if it will work, so yes if we store the id too will solve the problem i mentioned.

but, having set variable node, gives us more flexible in complex flows

like setting retries_count for some api calls "attempts = {{attempts}} + 1",
setting "user_type" to vip if some conditions met,
calculating cart total after each step by adding prices to {{total}} varriable and many more.

Saifallak:
edit: as far as i understand the code, duplicated btns ids doesn't matter, and it will work,

but still setVarriableNode will be good addition for complex flows,

i can see a good use case for me, is for displaying more options,

like getting categories in our DB, if they are 9+ the 10th button will be "show more"

and when clicked , will go to set varriable node, to increase "page" number and go back to same node that fetches from api.

shridarpatil:
Blocking (frontend assignments editor):

  1. Editing a variable name loses focus after one character. updateAssignmentKey rebuilds the object (delete old key, add new) on every keystroke, while the row is keyed by the variable name itself (:key="String(key)"), so the input remounts mid-type. Please key
  the v-for by index (or stage edits and commit on blur) so typing a name stays smooth. 
  2. Two blank rows collide. addAssignment inserts set[''] = ''; clicking "Add" twice produces two '' keys → duplicate :key and one row shadows the other. Same root cause as above — fixing the keying resolves it.

  Please address or document:

  3. Intra-node cross-references are order-dependent. The hint says you can reference other variables via {{...}}. Referencing a variable from an earlier node is fine, but referencing one set in the same node isn't reliable: execChatSetVariable iterates the set
  map with range (random order), and the set object doesn't preserve order on the wire either. So a = "X", b = "{{a}}" in one node may evaluate b first. Either change set to an ordered list of {key, value} pairs and process in order, or update the hint to say
  cross-references must point at variables set in earlier nodes.

  Non-blocking: 

  4. trigger_message / last_message become reserved keys — a flow that does store_as: "last_message" would be silently overwritten each turn. Worth a one-line note in the hint/docs.
  5. last_message only updates when userInput != "", so a WhatsApp Flow submission (data in flowResponseData, empty userInput) won't update it. Minor inconsistency, fine to leave.
  6. The added test covers trigger_message/last_message only — "comprehensive unit tests" is a bit generous since there's no test exercising the set_variable config→runner path. Not blocking if the backend node is already covered elsewhere.

PR #461, opened 2026-06-13
fix(chatbot): emit button:<id> edge conditions from flow builder

shridarpatil:
The buttons node's output handle used the bare button id, so edges drawn in the builder saved conditions without the button: prefix the graph runner matches — those button taps dead-ended. Prefix the handle id and self-heal bare conditions on load so existing flows re-save canonically.


Saifallak:
i have same issue locally, and fixed the same way, was going to make the same PR 😆

Issue #280, opened 2026-03-30
Unread message count issue


nilpatel42:
<img width="119" height="90" alt="Image" src="https://github.com/user-attachments/assets/48ff83e8-b6c5-482a-8b04-e71ca81ef094" />

When we have chat opened and received message then showing count of unread message in sidebar, to remove that count we need click on other chat and then again open that chat which have unread count

good to have like below in chat only not in sidebar

<img width="1131" height="207" alt="Image" src="https://github.com/user-attachments/assets/be6b428c-77d2-4d54-9670-9457230cbad2" />

PR #485, opened 2026-06-30
feat(notifications): OS desktop notifications + alert on unassigned messages

bambinounos:
## What
1. **Real OS-level desktop notifications** (Web Notification API) for new incoming messages when the Whatomate tab/window is backgrounded. No service worker is registered, so this uses the `Notification` constructor directly. Permission is requested on app load (`AppLayout.vue`), and Settings -> Notifications gets an explicit **"Enable desktop notifications"** button as a guaranteed user-gesture path (some browsers ignore the auto-request).
2. **Broaden the new-message alert** (sound + toast + desktop) to also fire for **unassigned / queue** messages, so nothing is missed when no agent is watching the chat (e.g. media-only first contacts). Messages assigned to a *different* agent stay silent.

## Why
`showNotification` currently only shows an in-app `toast.info`, so an agent working in another tab/app never sees new messages. And alerts only fire for messages already assigned to the current agent, so queue/unassigned messages are silent.

## Changes
- `services/websocket.ts`: `showDesktopNotification()` (fires only when permission granted and tab not focused; click focuses window + routes to chat); broadened the alert gate in `handleNewMessage`.
- `AppLayout.vue`: request notification permission on mount.
- `SettingsView.vue` + i18n (5 locales; en/es translated, ar/hi/ta English placeholders): explicit enable button + status.

No backend changes — the org-wide broadcast already carries `assigned_user_id`.

## Testing
Frontend `vite build` passes; full image builds and serves the embedded frontend.


shridarpatil:
Can you add a couple of screenshots of the notifications?

bambinounos:
Sorry for the delay — here they are, from a production deployment running this branch (Brave on Linux/GNOME).

**1. OS-level desktop notification.** Fires only while the Whatomate tab is backgrounded; clicking it focuses the window and routes to the conversation.

OS desktop notification

**2. The in-app alert that accompanies it**, with its View action. This is the part the second half of the PR broadens: it now also fires for unassigned/queue messages, so a first contact that lands with no agent watching is no longer silent.

In-app alert

The sound obviously doesn't survive a screenshot, but it fires on the same gate.

The second image is cropped just above the transfers table — that view lists real contacts and phone numbers, so this keeps them out of the thread rather than relying on a blur.

frontend/src/services/websocket.ts
```diff
@@ -320,33 +343,36 @@ class WebSocketService {
...
+    // the user's new_message_alerts setting.
     if (payload.direction === 'incoming' && !isViewingThisContact) {
       const authStore = useAuthStore()
       const currentUserId = authStore.user?.id
       const settings = authStore.userSettings
 
-      // Check if user is assigned to this contact
       const isAssignedToUser = payload.assigned_user_id === currentUserId
+      const isUnassigned = !payload.assigned_user_id // "" or undefined => queue
+      const shouldAlert = isAssignedToUser || isUnassigned
```
shridarpatil: **Alerting on unassigned messages leaks conversation content to users who can't access the conversation.**

`broadcastNewMessage` goes out via `WSHub.BroadcastToOrg` (`internal/handlers/messages.go:543`), and `Hub.broadcastMessage` (`internal/websocket/hub.go:149-168`) applies no permission filtering at all — every connected client in the org receives this payload. The `isAssignedToUser` check was the only thing keeping it off screen.

With `shouldAlert = isAssignedToUser || isUnassigned`:

- a user whose role lacks `chat` entirely (a campaign/billing/analytics-only role — see `frontend/src/components/layout/navigation.ts:61`), and
- an agent without `contacts:read`, who by design "only sees assigned chats" (`internal/handlers/contacts_test.go:1427-1443`)

now both get a toast **and an OS notification** containing the sender's profile name plus 50 characters of the customer's message body, for every unassigned conversation in the org. Secondary effect: clicking it routes them to `/chat/{contactId}` (line 58 above), which they can't open.

This needs a permission check before the unassigned branch — `authStore.hasPermission('chat', 'read')` at minimum. The better fix is server-side: have the backend target the broadcast at users who may see the conversation, rather than fanning out to the whole org and filtering in the client.

frontend/src/services/websocket.ts
```diff
@@ -320,33 +343,36 @@ class WebSocketService {
...
+    // messages). Messages assigned to a DIFFERENT agent stay silent. Respects
+    // the user's new_message_alerts setting.
     if (payload.direction === 'incoming' && !isViewingThisContact) {
       const authStore = useAuthStore()
       const currentUserId = authStore.user?.id
       const settings = authStore.userSettings
 
-      // Check if user is assigned to this contact
       const isAssignedToUser = payload.assigned_user_id === currentUserId
+      const isUnassigned = !payload.assigned_user_id // "" or undefined => queue
```
shridarpatil: **Chatbot traffic turns this into an alert storm.**

Conversations being handled by the chatbot are unassigned by definition, so every inbound customer turn in a bot conversation now fires `playNotificationSound()` + toast + desktop notification for *every* agent in the org, with no throttle or dedup.

That contradicts existing deliberate behaviour — `internal/handlers/messages.go:480-484` (`MarkIncomingRead`) explicitly quiets bot-handled chats: "Keeps the agent's contact-list unread count clean for conversations the bot is auto-handling. See issue #280."

An org with 50 concurrent bot conversations gives every agent a continuous notification sound. If the unassigned case is kept, it at least needs to exclude bot-handled conversations and be rate-limited.

frontend/src/services/websocket.ts
```diff
@@ -41,6 +41,29 @@ function showNotification(title: string, body: string, contactId: string) {
...
+// when permission is granted and the Whatomate tab/window is NOT focused, so an
+// agent working in another app/window still gets alerted. No service worker is
+// registered, so we use the Notification constructor directly. Falls back
+// silently when unsupported, denied, or the tab is already focused.
+let activeNotification: Notification | null = null
+function showDesktopNotification(title: string, body: string, contactId: string) {
+  if (typeof Notification === 'undefined' || Notification.permission !== 'granted') return
+  if (document.visibilityState === 'visible' && document.hasFocus()) return
+  try {
+    activeNotification?.close()
```
shridarpatil: This closes the previously shown notification regardless of which contact it belongs to, so at most one desktop notification can ever be visible.

Scenario: tab is backgrounded, contact A messages, contact B messages two seconds later — A's notification is dismissed before the agent ever looks at the screen, and that chat is silently gone from the notification tray.

The per-contact `tag: chat-${contactId}` on the next line already handles replacement for the *same* chat, which is the only case that should collapse. Suggest dropping this `close()` (or tracking `activeNotification` per contact id).

frontend/src/components/layout/AppLayout.vue
```diff
@@ -42,6 +42,14 @@ onMounted(() => {
         return null
       }
     })
+
+    // Request desktop-notification permission so agents get OS-level alerts for
+    // new messages even when the Whatomate tab is backgrounded. If a browser
+    // blocks the auto-request (needs a user gesture), Settings exposes an
+    // explicit "Enable desktop notifications" button as a guaranteed path.
+    if (typeof Notification !== 'undefined' && Notification.permission === 'default') {
```
shridarpatil: Requesting notification permission on app load with no user gesture is the pattern browsers actively penalise: Chrome treats repeated dismissals of an unprompted request as a signal to auto-block the origin, and Firefox rejects non-gesture requests outright.

The cost of landing on `denied` is high here — the Settings control (`SettingsView.vue:442`) then renders as a *disabled* "Blocked in browser" button with no instructions, so the user has no in-app recovery path. The auto-request can permanently cost the feature it's trying to enable.

The Settings button this PR already adds is the right trigger; I'd drop this block and rely on it alone. (The comment above says Settings is the fallback for when this is blocked — I'd invert that relationship.)

frontend/src/views/settings/SettingsView.vue
```diff
@@ -52,6 +52,23 @@ const notificationSettings = ref({
...
+// (some browsers ignore the auto-request on app load).
+const desktopPerm = ref<NotificationPermission>(
+  typeof Notification !== 'undefined' ? Notification.permission : 'denied'
+)
+async function enableDesktopNotifications() {
+  if (typeof Notification === 'undefined') {
+    toast.error(t('settings.desktopNotificationsUnsupported'))
+    return
+  }
+  desktopPerm.value = await Notification.requestPermission()
```
shridarpatil: `Notification.requestPermission()` has a legacy callback-only form (older WebKit / Safari < 16) that returns `undefined` rather than a promise.

There, this line sets `desktopPerm` to `undefined`: the permission dialog is shown and possibly granted, but the UI keeps showing "Enable" forever and the confirmation notification below never fires. The same call at `AppLayout.vue:51` calls `.catch()` on that `undefined`, throwing a `TypeError` out of the `onMounted` hook.

Cheap to guard either way — `Promise.resolve(Notification.requestPermission()).then(...)`, or just re-read `Notification.permission` after awaiting.

shridarpatil (commented):
Review of the desktop-notification work. The OS-notification plumbing itself is solid — the `typeof Notification` guards, the try/catch around the constructor, and the focus check all cover the right edge cases. My concerns are with broadening the alert gate to unassigned messages (comments 1 and 2), which I think needs to be reworked before this can land. Three smaller items follow.

On the locales: `pt.json` is the only one missing the new keys, but `fallbackLocale: 'en'` (`frontend/src/i18n/index.ts:70`) renders it identically to the English placeholders in `ar`/`hi`/`ta`, so no user-visible impact — not a blocker.

PR #489, opened 2026-06-30
fix(calls): two-way audio on incoming calls (silent caller)

bambinounos:
## Problem
On **incoming** WhatsApp calls the agent could not hear the caller — audio was one-way (the caller heard the agent, the agent heard silence). Two independent causes:

1. **Remote audio element garbage-collected.** `pc.ontrack` assigned the remote `MediaStream` to a local `new Audio()` that nothing retained, so it could be GC'd mid-call and go silent.
2. **Caller track attached too late.** The caller's media only starts *after* the agent answers, so `session.CallerRemoteTrack` is still nil when `completeTransferConnection` starts the `AudioBridge`. The caller→agent forward goroutine therefore never launched, and because `BridgeStarted` was already closed, the late `OnTrack` consumer returned immediately too — nobody forwarded caller→agent audio. (Confirmed in logs: `caller_remote_nil=true` with `Received remote track` arriving a beat later.)

## Fix
1. Hold the remote audio element on a stable store-scoped ref (released in `cleanup()`), so it isn't garbage-collected.
2. Add `AudioBridge.AttachCaller(callerRemote, agentLocal)` — idempotent, starts the late caller→agent forward goroutine and joins the bridge WaitGroup — and call it from the peer `OnTrack` handler when a transfer bridge is already active. The bridge becomes the track's sole reader, so inline-DTMF consumption is skipped in that case.

## Changes
- `frontend/src/stores/calling.ts`: keep `remoteAudioEl` referenced (both `ontrack` sites + cleanup).
- `internal/calling/bridge.go`: `AttachCaller` + guard fields.
- `internal/calling/webrtc.go`: wire the track into the live bridge on arrival.

## Testing
`go build ./internal/calling/...` + `go vet` pass; full image builds. Verified on a live deployment: incoming calls now have **two-way audio**; `outgoing` is unchanged.

### DIFF
--- a/frontend/src/stores/calling.ts
+++ b/frontend/src/stores/calling.ts
@@ -30,6 +30,9 @@ export const useCallingStore = defineStore('calling', () => {
   const callDuration = ref(0)
   const isMuted = ref(false)
   let durationTimer: number | null = null
+  // Remote (caller/consumer) audio element. Held on a stable ref so the browser
+  // doesn't garbage-collect it mid-call — otherwise the remote voice goes silent.
+  let remoteAudioEl: HTMLAudioElement | null = null
 
   // Call permission state (in-memory only, cleared on refresh)
   const callPermissions = reactive(new Map<string, { status: string, expiresAt?: string }>())
@@ -195,9 +198,9 @@ export const useCallingStore = defineStore('calling', () => {
 
     // Handle remote audio (caller's voice)
     pc.ontrack = (event) => {
-      const audio = new Audio()
-      audio.srcObject = event.streams[0]
-      audio.play().catch(() => { /* ignore autoplay */ })
+      if (!remoteAudioEl) remoteAudioEl = new Audio()
+      remoteAudioEl.srcObject = event.streams[0]
+      remoteAudioEl.play().catch(() => { /* ignore autoplay */ })
     }
 
     // Clean up when WebRTC connection drops
@@ -280,9 +283,9 @@ export const useCallingStore = defineStore('calling', () => {
 
     // Handle remote audio (consumer's voice)
     pc.ontrack = (event) => {
-      const audio = new Audio()
-      audio.srcObject = event.streams[0]
-      audio.play().catch(() => { /* ignore autoplay */ })
+      if (!remoteAudioEl) remoteAudioEl = new Audio()
+      remoteAudioEl.srcObject = event.streams[0]
+      remoteAudioEl.play().catch(() => { /* ignore autoplay */ })
     }
 
     // Clean up when WebRTC connection drops
@@ -418,6 +421,11 @@ export const useCallingStore = defineStore('calling', () => {
     if (localStream.value) {
       localStream.value.getTracks().forEach(t => t.stop())
       localStream.value = null
+    }
+    if (remoteAudioEl) {
+      remoteAudioEl.pause()
+      remoteAudioEl.srcObject = null
+      remoteAudioEl = null
     }
     isOnCall.value = false
     isOnHold.value = false
--- a/frontend/src/stores/calling.ts
+++ b/frontend/src/stores/calling.ts
@@ -198,6 +198,10 @@ export const useCallingStore = defineStore('calling', () => {
 
     // Handle remote audio (caller's voice)
     pc.ontrack = (event) => {
+      // A queued ontrack can still fire after cleanup() tore this call down
+      // (or after a new call replaced the connection); recreating the audio
+      // element here would leak it and play ghost audio from a dead stream.
+      if (peerConnection.value !== pc) return
       if (!remoteAudioEl) remoteAudioEl = new Audio()
       remoteAudioEl.srcObject = event.streams[0]
       remoteAudioEl.play().catch(() => { /* ignore autoplay */ })
@@ -283,6 +287,9 @@ export const useCallingStore = defineStore('calling', () => {
 
     // Handle remote audio (consumer's voice)
     pc.ontrack = (event) => {
+      // Same late-ontrack guard as in acceptTransfer: never re-create
+      // the audio element for a connection that is no longer the active one.
+      if (peerConnection.value !== pc) return
       if (!remoteAudioEl) remoteAudioEl = new Audio()
       remoteAudioEl.srcObject = event.streams[0]
       remoteAudioEl.play().catch(() => { /* ignore autoplay */ })
--- a/internal/calling/bridge.go
+++ b/internal/calling/bridge.go
@@ -30,6 +30,11 @@ type AudioBridge struct {
 	firstAgentSeq bool // true until the first agent packet sets the base
 	agentBaseSeq  uint16
 	agentBaseTS   uint32
+
+	// callerAttached guards AttachCaller so the late caller→agent forwarding
+	// goroutine (used when the caller's track arrives after Start) runs once.
+	mu             sync.Mutex
+	callerAttached bool
 }
 
 // NewAudioBridge creates a new audio bridge with optional per-direction recorders.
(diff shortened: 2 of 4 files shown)


internal/calling/bridge.go
```diff
@@ -72,6 +77,34 @@ func (b *AudioBridge) Start(
...
+	}
+	b.mu.Lock()
+	if b.callerAttached {
+		b.mu.Unlock()
+		return
+	}
+	b.callerAttached = true
+	b.mu.Unlock()
+
+	b.wg.Add(1)
```
shridarpatil: **`wg.Add(1)` here races `Start`'s `wg.Wait()` and can panic the process.** The doc comment says this is safe "because the agent→caller goroutine keeps the counter positive" — but that only holds while that goroutine is alive. In the exact scenario this PR targets, `Start` was entered with `callerRemote == nil`, so the agent→caller goroutine is the *only* one and the counter is 1. Two triggers break it:

- **Natural exit:** the agent leg's `src.Read` errors (agent hangs up / network blip) → `defer wg.Done()` drops the counter 1→0 and `wg.Wait()` returns, at the same instant `OnTrack` calls `AttachCaller` → `wg.Add(1)`. The `select { case <-b.stop }` guard does *not* help — `b.stop` is closed only by `Stop()`, never on a read-error exit.
- **Teardown TOCTOU:** the `<-b.stop` select passes (still open), then `EndTransfer`/`HoldCall` runs `bridge.Stop(); bridge.Wait()` before this line runs.

Either way `Add` runs concurrent with the final `Wait` decrement → `panic: sync: WaitGroup is reused before previous Wait has returned`. `completeTransferConnection` runs in its own goroutine with no `recover`, so this crashes the process.

Consider having `Start` always reserve a caller slot it can later feed (single owner of `wg.Add`), or add the caller direction inside the bridge under `b.mu` gated on an explicit "stopped" flag, rather than `Add`-ing from outside after `Wait` may have begun.

shridarpatil (commented):
Automated review — the fix is the right shape and was verified on a live deployment. A few concurrency/lifecycle edge cases the change opens up are noted inline. Findings 1 and 4 are the two I'd want addressed before merge; 2 and 3 are narrower.

internal/calling/bridge.go
```diff
@@ -72,6 +77,34 @@ func (b *AudioBridge) Start(
...
+	if callerRemote == nil || agentLocal == nil {
+		return
+	}
+	select {
+	case <-b.stop:
+		return
+	default:
+	}
+	b.mu.Lock()
+	if b.callerAttached {
```
shridarpatil: **`callerAttached` guards two `AttachCaller` calls against each other, but not `AttachCaller` vs `Start`.** On transfer/resume paths `Start` gets a non-nil `callerRemote` and launches a caller→agent forward *without* setting `callerAttached`. If `OnTrack` fires again for the caller audio track (ICE renegotiation / restart), `AttachCaller` sees `callerAttached == false` and launches a **second** forward on the same `*TrackRemote`: two goroutines `Read` one track (RTP split arbitrarily → corrupt/halved agent audio) and both call `b.callerRec.WritePacket` concurrently (data race → corrupt OGG recording). Low likelihood in the single-m-line WhatsApp flow, but unhandled — worth setting `callerAttached` from `Start` too when it launches the caller direction.

internal/calling/webrtc.go
```diff
@@ -66,8 +66,19 @@ func (m *Manager) negotiateWebRTC(session *CallSession, account *models.WhatsApp
...
 		session.CallerRemoteTrack = track
+		bridge := session.Bridge
+		agentLocal := session.AgentAudioTrack
 		session.mu.Unlock()
 
+		// On incoming calls the caller's media often starts only after the agent
+		// answers, so the transfer bridge may already be running without the
+		// caller track (one-way audio). Wire the track into the live bridge now
+		// so the agent can hear the caller; the bridge becomes the sole reader.
+		if bridge != nil && agentLocal != nil {
```
shridarpatil, line 77: **Residual undrained-track window — narrows but doesn't fully close the original bug.** `completeTransferConnection` snapshots `callerRemote = session.CallerRemoteTrack` (nil) *before* `setupAudioBridge` assigns `session.Bridge`. If the caller track arrives in that gap, `OnTrack` reads `bridge == nil` here, falls through to `consumeAudioWithDTMF`, which no-ops as soon as `BridgeStarted` closes — and since `Start` got `callerRemote == nil`, nobody ever drains the caller track → one-way audio again (pion receive-buffer backpressure). Much narrower than before, but the same failure mode. A re-check after the bridge is known-started (or attaching off `session.CallerRemoteTrack` once the bridge exists) would close it.

frontend/src/stores/calling.ts
```diff
@@ -195,9 +198,9 @@ export const useCallingStore = defineStore('calling', () => {
 
     // Handle remote audio (caller's voice)
     pc.ontrack = (event) => {
-      const audio = new Audio()
-      audio.srcObject = event.streams[0]
-      audio.play().catch(() => { /* ignore autoplay */ })
+      if (!remoteAudioEl) remoteAudioEl = new Audio()
```
shridarpatil, line 205: **A late/queued `ontrack` firing after `cleanup()` re-materializes an orphan `remoteAudioEl`.** A WS event (`call_ended`, `outgoing_call_ended`, `call_transfer_completed`) runs `cleanup()` — pausing/nulling `remoteAudioEl` — while an `ontrack` from `setRemoteDescription` is still queued. The queued handler then hits `if (!remoteAudioEl) remoteAudioEl = new Audio()`, allocates a fresh element with the stale stream, and plays it. Because `cleanup()` already ran, nothing reclaims it until the *next* call's cleanup → brief ghost audio after hangup plus a leaked element/`MediaStream`. Ironically the old `const audio = new Audio()` was auto-GC'd here; the intentional stable ref is what makes it linger. Guard the handler on `peerConnection.value` / call-active state before (re)creating the element.

internal/calling/bridge.go
bambinounos: Fixed in f56a5bd. You're right — the "counter stays positive" claim only held while the agent leg was alive, and neither the natural-exit path nor the Stop/Wait TOCTOU was covered.

Went with your first suggestion: `Start` is now the **single owner of `wg.Add`** (every `Add` happens inside `Start`, before its `Wait`). When it can't launch the caller leg directly, it reserves a *slot goroutine* that waits on either `b.stop` or a buffered (cap 1) `callerSlot` channel. `AttachCaller` just feeds that channel — it never touches the WaitGroup — so it can no longer race `Wait` regardless of when the other legs exit. `Stop()` also sets an explicit `stopped` flag under `b.mu`, making `AttachCaller` a guaranteed no-op after teardown begins (if the flag check races the close, the leg value is either discarded unread or `forward` exits on its first stop-channel check).

One deliberate semantic change: with the slot reserved, `Start` now blocks until `Stop()` when the caller leg never materializes (before, it returned as soon as the agent leg died). All call sites run `Start` at the tail of a dedicated goroutine and every teardown path (`EndTransfer`, `HoldCall`, session cleanup) calls `Stop()`, so this just parks one goroutine until teardown instead of leaving an unattachable half-dead bridge.

internal/calling/bridge.go
bambinounos: Fixed in f56a5bd — `Start` now sets `callerAttached` under `b.mu` whenever it launches the caller→agent leg directly, so a caller-track `OnTrack` refire goes through the idempotency check and can't spawn a second reader on the same `*TrackRemote` (no split RTP, no concurrent `WritePacket` on the recorder).

internal/calling/webrtc.go
bambinounos: Fixed in f56a5bd. Since `OnTrack` stores `session.CallerRemoteTrack` and reads `session.Bridge` under the same `session.mu` that `setupAudioBridge` uses to assign the bridge, the two sides now form a complete hand-off:

- track lands **after** the bridge is assigned → `OnTrack` sees it and calls `AttachCaller` (existing path);
- track lands **before/during** the gap → `completeTransferConnection` re-reads the direction-appropriate remote track (`lateCallerTrack`) after `safeClose(BridgeStarted)` and, if its earlier snapshot was nil, hands the track to `Start` directly.

Placed after `BridgeStarted` closes so `consumeAudioWithDTMF` has been told to stand down first — no window with two readers. Applied the same re-check to `ResumeCall`, which has the identical snapshot→assignment shape (relevant when a call is held before the caller's media ever arrived). `AttachCaller`'s idempotency covers the both-sides-fire overlap.

frontend/src/stores/calling.ts
bambinounos: Fixed in f56a5bd — both `ontrack` handlers (transfer accept + outgoing call) now guard on `peerConnection.value !== pc` before touching `remoteAudioEl`. `cleanup()` nulls `peerConnection`, so a queued `ontrack` firing after teardown returns early instead of re-materializing the element with the dead stream; it also covers the case where a *new* call has already replaced the connection (the stale pc's late track can't hijack the fresh call's audio element). Identity comparison is safe here — Vue 3 leaves non-plain objects like `RTCPeerConnection` un-proxied inside refs.

bambinounos:
Thanks for the thorough review — all four findings are addressed in f56a5bd (1 and 4 as requested, plus 2 and 3 since they were cheap once the bridge owned its lifecycle properly):

1. **WaitGroup race** → `Start` is now the single owner of `wg.Add`; when the caller leg can't start yet it reserves a slot goroutine that `AttachCaller` feeds via a buffered channel, so `Add` can never race `Wait`. `Stop()` sets an explicit `stopped` flag under the mutex.
2. **`callerAttached` vs `Start`** → `Start` marks the flag when it launches the caller leg directly.
3. **Residual undrained-track window** → `completeTransferConnection` (and `ResumeCall`, same shape) re-checks the session's caller track once the bridge is visible; combined with `OnTrack`'s existing attach-if-bridge path under the same `session.mu`, the hand-off now has no gap.
4. **Late `ontrack` after `cleanup()`** → both handlers ignore events whose `pc` is no longer the active `peerConnection`.

`go build ./...` + `go vet` and the frontend build pass. This is running on our production deployment as before; happy to adjust anything further.

PR #493, opened 2026-07-02
feat(webhooks): custom body templates for outbound webhooks

bamdadd:
## What

Adds an optional per-webhook **body template** that lets an outbound webhook render any payload shape — so you can point a webhook straight at a Slack / Discord / Teams Incoming Webhook without a middleware proxy.

## Why

Outbound webhooks always sent a fixed envelope:

```json
{ "event": "message.incoming", "timestamp": "...", "data": { ... } }
```

Chat platforms reject this — a Slack Incoming Webhook requires `{"text": "..."}` (or blocks) and returns `invalid_payload` for the envelope. There was no way to reshape the body, so integrating Slack/Discord/Teams required an external transformer.

## How

- **Model**: new `Webhook.BodyTemplate` (`text` column, added by AutoMigrate). Empty preserves the current envelope — fully backward compatible.
- **Rendering** (`renderWebhookBody`): with no template, marshals the envelope as before; with a template, renders a Go `text/template` given:
  - `.Event` — event name
  - `.Timestamp` — RFC3339 timestamp
  - `.Data.*` — event payload keyed by **JSON field names** (`.Data.contact_name`, `.Data.content`, …)
  - a `json` helper that marshals a value to a properly quoted/escaped JSON literal, so interpolated content (quotes, newlines) stays valid JSON. `upper`/`lower` also provided.
- **Validation**: templates are compiled on create/update — a syntax error returns `400` instead of silently failing at delivery time.
- **Test button**: renders the template against a sample message so you can confirm the target accepts the shape before going live.
- **UI**: a body-template textarea on the webhook detail page with a one-click "Use Slack example" prefill and inline help.

### Slack example

```go
{"text": {{ json (printf "*%s* from %s:\n%s" .Event .Data.contact_name .Data.content) }}}
```

→

```json
{"text": "*message.incoming* from Jane:\nHello!"}
```

## Tests

- `webhook_template_test.go`: empty→envelope, Slack template → valid JSON, `json` helper escaping (quotes/newlines), event/timestamp exposure, template validation (valid + malformed), and `dataToMap` field-name keying. All passing; `go vet` clean.
- Frontend `typecheck` + `eslint` clean on changed files.
- Docs: outbound-webhook + body-template section added to `webhooks.mdx`.


bamdadd:
This will allow me to do slack or any other api integration without manually writing integrations for them

PR #494, opened 2026-07-02
fix(ws): reconnect forever with capped backoff and reconnect on device wake

bambinounos:
## Problem

The WebSocket client gives up after 5 reconnect attempts (`maxReconnectAttempts`). On mobile browsers this leaves the app permanently disconnected: Android freezes background tabs (timers throttled, socket killed), so by the time the agent unlocks their phone the 5 attempts are long exhausted. The tab then shows stale data until a manual refresh — easy to miss, and it looks like "messages stopped arriving".

Even on desktop, a laptop resuming from suspend can burn through the 5 attempts before the network is back.

## Fix

- **Never give up reconnecting.** The retry cap is replaced with an `intentionalClose` flag, so `disconnect()` (logout) still stops reconnection, while transient failures retry forever with exponential backoff **capped at 30s** (exponent also clamped to avoid overflow).
- **Reconnect immediately on wake.** New lifecycle listeners (`visibilitychange` → visible, `online`, `pageshow`) reconnect on the spot when the socket is dead instead of waiting for the next backoff tick, resetting the backoff counter. The existing `refreshStaleData()` already resyncs contacts/transfers after reconnect, so recovery is complete.

## Testing

- Airplane mode for 2+ minutes → back online: reconnects instantly ("Connection restored" toast), no stuck state.
- Phone locked/backgrounded for extended periods → unlock: reconnects on `visibilitychange`.
- Logout: no reconnect attempts (flag respected).
- Verified in production on a multi-agent deployment (desktop + Android).


frontend/src/services/websocket.ts
```diff
@@ -534,18 +539,53 @@ class WebSocketService {
...
+    this.lifecycleListenersInstalled = true
+
+    const reconnectIfDead = () => {
+      if (this.intentionalClose) {
+        return
+      }
+      const state = this.ws?.readyState
+      if (state !== WebSocket.OPEN && state !== WebSocket.CONNECTING) {
+        this.reconnectAttempts = 0
+        this.connect()
```
shridarpatil: `connect()` only short-circuits on `readyState === OPEN` (line 110), and it `await`s `getTokenFn()` before assigning `this.ws`. During that async token-fetch window `this.ws` still holds the old CLOSED socket, so neither the OPEN nor the CONNECTING guard here catches it.

On a mobile wake, `online` + `visibilitychange` (+ `pageshow`) fire near-simultaneously while a backoff `setTimeout(connect, delay)` may already be pending. Multiple of these can call `connect()` before any of them assigns `this.ws`, so each reaches `new WebSocket(url)` and reassigns `this.ws` — leaving orphaned sockets whose `onclose` handlers each call `handleReconnect()`. A single wake can then fork the retry into several concurrent backoff chains and open duplicate connections, i.e. a self-inflicted reconnect storm in exactly the flaky-network scenario this PR targets.

Suggest an in-flight guard: set an `isConnecting` flag at the top of `connect()`, include `CONNECTING` in the early-return check, and clear the flag in both `onopen` and `onclose`.

frontend/src/services/websocket.ts
```diff
@@ -534,18 +539,53 @@ class WebSocketService {
...
+    }
+    this.lifecycleListenersInstalled = true
+
+    const reconnectIfDead = () => {
+      if (this.intentionalClose) {
+        return
+      }
+      const state = this.ws?.readyState
+      if (state !== WebSocket.OPEN && state !== WebSocket.CONNECTING) {
+        this.reconnectAttempts = 0
```
shridarpatil: Minor / lower confidence: `reconnectIfDead` resets `reconnectAttempts = 0` and calls `connect()` directly rather than going through `handleReconnect()`. If `getTokenFn()` returns `null` (e.g. auth expired), `connect()` returns early without creating a socket, so no `onclose` fires and no retry is scheduled — the "never give up" chain silently terminates until the next `visibilitychange`/`online`/`pageshow` event.

This is pre-existing behavior (the null-token early return never scheduled a retry), but the PR's indefinite-retry promise makes it more noticeable. Non-blocking — worth a comment or a fallback `handleReconnect()` on null token.

frontend/src/services/websocket.ts
bambinounos: Good catch — fixed in 73b096a exactly as suggested: an `isConnecting` flag is set at the top of `connect()` and covers the async token-fetch window, and the early-return now also checks `readyState === CONNECTING`. The flag is cleared in `onopen`, `onclose`, the null-token return, and the constructor `catch`, so a failed attempt can never leave the guard stuck. With this, simultaneous wake events (visibilitychange/online/pageshow + a pending backoff timer) collapse into a single connection attempt.

frontend/src/services/websocket.ts
bambinounos: Also addressed in 73b096a: the null-token path now calls `handleReconnect()` before returning (with a comment explaining why — no socket means no `onclose`, so nothing else would reschedule). Since `handleReconnect()` respects `intentionalClose`, logout still stops the chain, and the capped backoff keeps it cheap if auth has genuinely expired until the user re-authenticates.

bambinounos:
Thanks for the review — both findings addressed in 73b096a: (1) `isConnecting` flag + `CONNECTING` in the early-return so concurrent wake events/backoff timers can't open duplicate sockets, cleared on every exit path; (2) null-token path now schedules `handleReconnect()` so the retry chain survives a transient token-fetch failure. Frontend build passes.

frontend/src/services/websocket.ts
```diff
@@ -117,6 +134,11 @@ class WebSocketService {
     // Get a fresh short-lived WS token
     const token = this.getTokenFn ? await this.getTokenFn() : null
     if (!token) {
+      // No socket was created, so no onclose will ever fire: schedule the
+      // retry here or the reconnect chain would silently end on a transient
+      // token-fetch failure (backoff keeps this cheap if auth truly expired).
+      this.isConnecting = false
+      this.handleReconnect()
```
shridarpatil, line 150: 🔴 **Blocking.** This is the heart of the post-logout infinite loop. After logout, `getWSToken()` 401s → `token` is `null` → this reschedules `handleReconnect()`. Since `disconnect()` is never called (see line 192), `intentionalClose` is `false`, so it loops forever at ≤30s intervals — a logged-out tab polling a 401 indefinitely.

Distinguish "auth genuinely gone" from "transient token-fetch failure": skip rescheduling here when `!useAuthStore().isAuthenticated`.

frontend/src/services/websocket.ts
```diff
@@ -158,18 +182,19 @@ class WebSocketService {
...
       }
     } catch {
+      this.isConnecting = false
       this.handleReconnect()
     }
   }
 
   disconnect() {
     this.stopPing()
+    this.intentionalClose = true // Prevent reconnect (deliberate close, e.g. logout)
```
shridarpatil, line 202: The whole `intentionalClose` stop-mechanism hinges on `disconnect()` being called — but **nothing in the frontend calls `disconnect()`**. Logout is `authStore.logout()` → `clearAuth()` → `router.push('/login')`, which never touches the WS service. So `intentionalClose` stays `false` and reconnection is never actually stopped.

Call `wsService.disconnect()` from `clearAuth()`/logout to make this real.

frontend/src/services/websocket.ts
```diff
@@ -534,18 +559,53 @@ class WebSocketService {
...
   private handleReconnect() {
-    if (this.reconnectAttempts >= this.maxReconnectAttempts) {
+    if (this.intentionalClose) {
       return
     }
 
+    // Never give up: mobile browsers freeze background tabs for arbitrarily
+    // long, so a retry cap would leave the app permanently disconnected.
+    // Exponential backoff capped at maxReconnectDelay keeps retries cheap.
     this.reconnectAttempts++
```
shridarpatil, line 601: Removing the attempt cap is the intended change, but it's only safe if reconnection reliably stops on logout — which it currently doesn't (line 192). With the cap gone and `intentionalClose` never set, this becomes an unbounded loop for logged-out tabs. Please add an auth gate here too: `if (this.intentionalClose || !useAuthStore().isAuthenticated) return`.

frontend/src/services/websocket.ts
```diff
@@ -534,18 +559,53 @@ class WebSocketService {
...
-    const delay = this.reconnectDelay * Math.pow(2, this.reconnectAttempts - 1)
+    const delay = Math.min(
+      this.reconnectDelay * Math.pow(2, Math.min(this.reconnectAttempts - 1, 10)),
+      this.maxReconnectDelay
+    )
 
     setTimeout(() => {
```
shridarpatil: 🟠 The timer handle isn't stored, so `disconnect()` can't cancel it. Even once `disconnect()` is wired up, a backoff timer scheduled just before it will still fire `connect()`, which resets `intentionalClose = false` and reopens the socket. Save the id to a field and `clearTimeout` it in `disconnect()`.

shridarpatil (changes requested):
The core idea is sound, and the wake handling (`visibilitychange`/`online`/`pageshow`) plus the `isConnecting` guard against concurrent `connect()` during the async token fetch are correct and worthwhile. But there's one blocking bug that inverts the PR's own goal.

## 🔴 Blocking: infinite reconnect loop after logout

The stop-mechanism is `disconnect()` setting `intentionalClose = true` ("Logout: no reconnect attempts (flag respected)"). **`disconnect()` is dead code — nothing in the frontend ever calls it.** Logout runs `authStore.logout()` → `clearAuth()` → `router.push('/login')`, a client-side nav (not a reload), so the singleton `wsService` and its `setTimeout` chain survive, and `intentionalClose` stays `false` forever.

Combined with this PR (a) removing the 5-attempt cap and (b) making the null-token branch call `handleReconnect()` instead of returning, once the socket closes after logout:

`onclose` → `handleReconnect()` → `connect()` → `getWSToken()` returns 401→`null` → **reschedules `handleReconnect()`** → … forever, at ≤30s intervals. Every logged-out-but-open tab now polls the WS-token endpoint with a 401 indefinitely, and the new lifecycle listeners re-trigger it on every device wake. Previously this was bounded (null-token returned without rescheduling; the 5-attempt cap capped `onclose` retries).

**Fix:** call `wsService.disconnect()` in `clearAuth()`/logout so the `intentionalClose` design actually runs, **and** gate reconnection on auth (bail in `handleReconnect`/`reconnectIfDead` when `!isAuthenticated`, or skip rescheduling on null token when unauthenticated — distinguish "auth gone" from "transient fetch failure").

## 🟠 Secondary: `disconnect()` can't cancel an in-flight backoff

Even once `disconnect()` is wired up, the backoff `setTimeout` handle is never stored, so a timer scheduled before `disconnect()` still fires `connect()` afterward — and `connect()` unconditionally resets `intentionalClose = false`, reopening the socket. Store the timeout id and `clearTimeout` it in `disconnect()`.

## Minor (non-blocking)
- Lifecycle listeners use anonymous handlers and can never be removed; acceptable for a singleton, but it's why post-logout wake-retry keeps firing — the auth gate above neutralizes it.
- The inner `Math.min(reconnectAttempts - 1, 10)` clamp is redundant given the outer `Math.min(..., maxReconnectDelay)`, but harmless.

Good direction — requesting changes for the logout loop.

bambinounos:
You're right on all counts — fixed in ad26f55. The blocking loop was real and, embarrassingly, my own null-token change unmasked it (previously the null-token return + the 5-attempt cap kept it bounded).

**🔴 Logout loop.** Confirmed `disconnect()` was dead code and logout is a client-side nav (`AppLayout` → `authStore.logout()` → `clearAuth()` → `router.push('/login')`), so the singleton and its backoff `setTimeout` survived with `intentionalClose = false`. Two-layer fix, as you suggested:
- `clearAuth()` now calls `wsService.disconnect()`, so the `intentionalClose` design actually runs — and it covers the invalid-session clears in `restoreSession()` too, not just the logout button.
- `handleReconnect()` and `reconnectIfDead()` now also bail on `!isAuthenticated()`. That's the "distinguish auth-gone from transient fetch failure" gate: a null token while still logged in keeps retrying (backoff), a null token after the session is gone stops. `isAuthenticated()` reads the auth store lazily and is wrapped in try/catch so a pre-Pinia call is treated as unauthenticated (bail).

**🟠 In-flight backoff.** The pending timer is now tracked in `reconnectTimer` and `disconnect()` `clearTimeout`s it (also clearing `isConnecting`), so a timer scheduled just before `disconnect()` can no longer fire `connect()` and flip `intentionalClose` back to false. `handleReconnect()` also clears any prior timer before scheduling, so overlapping triggers can't stack chains.

**Minor.** The anonymous lifecycle listeners stay (singleton), now neutralized by the auth gate as you noted. I left the redundant inner `Math.min(..., 10)` clamp to keep the diff focused on the fix.

Frontend build passes. Thanks for catching this — it genuinely inverted the PR's goal.

Issue #100, opened 2026-01-29
Cannot re-add member to team after removal - "Failed to add member" error


alwyndon05:
When a team member is removed from a team, attempting to add them back results in a "Failed to add member" error. The user cannot be re-added to the team they were previously removed from.

Steps to reproduce the behaviour:

- Go to Settings > Teams 
- Click on the members icon for any team (e.g., "ABC Team")
- Add a member to the team by clicking "Add as Agent" or "Add as Manager"
- Remove the member from the team by clicking the remove icon
- Try to add the same member back to the team
- See error: "Failed to add member"

Expected behavior
After removing a member from a team, the user should be able to re-add them to the same team without any errors. The add operation should work the same as it did when the user was first added.

Additional context

The "Member removed from team" success message appears, suggesting the backend removal works
The removed user reappears in the "Add Members" list, indicating the UI recognises they're no longer a member

However, the subsequent add operation fails, suggesting there may be:

- A soft-delete issue where the membership record still exists in a deleted state
- A unique constraint violation in the database
- Missing cleanup of the previous membership record
- A race condition or caching issue

shridarpatil:
Added a fix for this. `DROP INDEX IF EXISTS idx_team_members_unique` drop existing index from the database, and re-run migration new index will get added.

PR #499, opened 2026-07-03
feat(chat): forward messages to another contact

bambinounos:
## Why

The Cloud API has no native forward endpoint — the `forwarded` flag exists only as inbound metadata, and the "Forwarded" label is consumer-app only. Agents coming from other inboxes expect to forward a message (an address, a photo, a document) to another contact without re-typing or re-downloading it, so this implements the same approach commercial inboxes use: **re-send the content as a new message**.

## What

**Backend** — `POST /api/messages/{id}/forward` with `{contact_id, whatsapp_account?}` (new `internal/handlers/messages_forward.go`):
- The agent must have access to **both** conversations (same assignment scoping as the rest of the chat handlers); forwarding to the same conversation is rejected.
- Outgoing account resolution: request override → destination contact's account → default account.
- `text` copies the content; `image/video/audio/document` re-read the file from local storage (reusing `ServeMedia`'s path-traversal/symlink guards) and re-upload it, since inbound Meta media IDs are **not reusable for sending**. Captions are preserved. Other types → 400.
- Delivery goes through `SendOutgoingMessage`, so status tracking, WS broadcast and webhooks behave like any outgoing message — a closed 24h service window on the destination surfaces as a normal failed message with its `error_message`.

**Frontend** — a Forward button on message hover next to Reply (both directions, only for forwardable types) opening a dialog with debounced server-side contact search (300 ms, excludes the current contact). i18n: en/es translated; ar/hi/ta get English placeholders (same pattern as the existing locales).

## Limitations (platform, not implementation)

- The recipient sees a normal message — no "Forwarded" label is possible through the Cloud API.
- View-once messages can't be forwarded (WhatsApp blocks this by design across the platform).

## Testing

- `go build ./...` + `go vet`, frontend build pass.
- Verified on our production deployment: forwarding text, images and documents between real conversations (media re-uploads correctly, captions preserved); non-forwardable types don't show the button.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/frontend/src/views/chat/ChatView.vue
+++ b/frontend/src/views/chat/ChatView.vue
@@ -87,7 +87,8 @@ import {
   Code,
   RotateCw,
   Filter,
-  StickyNote
+  StickyNote,
+  Forward
 } from 'lucide-vue-next'
 import { getInitials, getAvatarGradient } from '@/lib/utils'
 import { useColorMode } from '@/composables/useColorMode'
@@ -1179,6 +1180,67 @@ function replyToMessage(message: Message) {
   nextTick(() => {
     messageInputRef.value?.focus()
   })
+}
+
+// Forward message: the Cloud API has no native forward, so the backend
+// re-sends the content to the chosen contact as a new message.
+const isForwardDialogOpen = ref(false)
+const forwardingMessage = ref<Message | null>(null)
+const forwardSearchQuery = ref('')
+const forwardResults = ref<any[]>([])
+const isSearchingForward = ref(false)
+const forwardSendingId = ref<string | null>(null)
+let forwardSearchTimer: number | undefined
+
+const forwardableTypes = ['text', 'image', 'video', 'audio', 'document']
+function canForward(message: Message): boolean {
+  return forwardableTypes.includes(message.message_type)
+}
+
+function openForwardDialog(message: Message) {
+  forwardingMessage.value = message
+  forwardSearchQuery.value = ''
+  isForwardDialogOpen.value = true
+  searchForwardContacts()
+}
+
+async function searchForwardContacts() {
+  isSearchingForward.value = true
+  try {
+    const response = await contactsService.list({
+      search: forwardSearchQuery.value.trim() || undefined,
+      limit: 20
+    })
+    const data = response.data.data || response.data
+    forwardResults.value = (data.contacts || []).filter(
+      (c: any) => c.id !== contactsStore.currentContact?.id
+    )
+  } catch {
+    forwardResults.value = []
+  } finally {
+    isSearchingForward.value = false
+  }
+}
+
+watch(forwardSearchQuery, () => {
+  if (!isForwardDialogOpen.value) return
+  window.clearTimeout(forwardSearchTimer)
+  forwardSearchTimer = window.setTimeout(searchForwardContacts, 300)
+})
+
+async function forwardToContact(contact: any) {
+  if (!forwardingMessage.value || forwardSendingId.value) return
+  forwardSendingId.value = contact.id
+  try {
+    await messagesService.forward(forwardingMessage.value.id, contact.id)
+    toast.success(t('chat.forwarded'))
+    isForwardDialogOpen.value = false
+    forwardingMessage.value = null
+  } catch (error: any) {
+    toast.error(error?.response?.data?.message || t('chat.forwardFailed'))
+  } finally {
+    forwardSendingId.value = null
+  }
 }
 
 // Watch for slash commands in message input
@@ -2317,6 +2379,16 @@ async function sendMediaMessage() {
                 >
                   <Reply class="h-3 w-3" />
                 </Button>
+                <Button
+                  v-if="canForward(message)"
+                  variant="ghost"
+                  size="icon"
+                  class="h-6 w-6"
+                  :title="$t('chat.forward')"
+                  @click="openForwardDialog(message)"
+                >
+                  <Forward class="h-3 w-3" />
+                </Button>
               </div>
               <!-- Reply button for outgoing messages (shown on hover) -->
               <div v-if="message.direction === 'outgoing'" class="flex flex-col gap-0.5 opacity-0 group-hover:opacity-100 transition-opacity self-center ml-1">
@@ -2346,6 +2418,16 @@ async function sendMediaMessage() {
                   @click="replyToMessage(message)"
                 >
                   <Reply class="h-3 w-3" />
+                </Button>
+                <Button
+                  v-if="canForward(message)"
+                  variant="ghost"
+                  size="icon"
+                  class="h-6 w-6"
+                  :title="$t('chat.forward')"
+                  @click="openForwardDialog(message)"
+                >
+                  <Forward class="h-3 w-3" />
                 </Button>
                 <Button
                   v-if="message.status === 'failed' && message.message_type !== 'template'"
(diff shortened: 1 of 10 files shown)


bambinounos:
Field-testing follow-up (pushed 786c11b): forwarding an inbound **image** (or video/audio/sticker) failed with Meta `(#100) The parameter file is required`. Those inbound types are stored without a `MediaFilename` (only documents carry one), so `UploadMedia` built a multipart part with `filename=""`, which the Cloud API treats as a plain field instead of a file upload — forwarding documents worked, images didn't.

Fixed two ways: `ForwardMessage` now synthesizes a filename from the mime type when the source has none, and `UploadMedia` defensively defaults an empty filename to `"file"` so no caller can emit the invalid empty-filename part. Verified forwarding images end to end in production.

pkg/whatsapp/client.go
```diff
@@ -303,6 +303,14 @@ type UploadMediaResponse struct {
 func (c *Client) UploadMedia(ctx context.Context, account *Account, data []byte, mimeType, filename string) (string, error) {
 	url := fmt.Sprintf("%s/%s/%s/media", c.getBaseURL(), account.APIVersion, account.PhoneID)
 
+	// An empty multipart filename makes Meta treat the part as a plain field
+	// rather than a file upload, failing with "(#100) The parameter file is
+	// required". Guarantee a non-empty name for any caller (e.g. forwarding
+	// inbound images, which carry no filename).
+	if filename == "" {
+		filename = "file"
```
shridarpatil, line 314: **medium** — actual defect is at `pkg/whatsapp/client.go:324` (and `:325`), just outside this hunk: `filename` and `mimeType` are `Fprintf`'d raw into the `Content-Disposition` / `Content-Type` multipart headers.

Forwarding is the first caller that feeds *remote-contact-controlled* data there — `message.MediaFilename` comes from `msg.Document.Filename` in the inbound webhook (`internal/handlers/chatbot_processor.go:1453`) and is stored unsanitized. A contact who sends a document named `a"; name="messaging_product` — or one containing CRLF — corrupts the multipart body when an agent forwards it, breaking the upload or injecting extra parts/headers into the request to Meta.

Since this hunk already touches filename handling, switch the body to `mime/multipart.Writer` (which escapes `CreateFormFile` names) or at minimum strip `"`, CR and LF from `filename` here.

internal/handlers/messages_forward.go
```diff
@@ -0,0 +1,167 @@
...
+	switch message.MessageType {
+	case models.MessageTypeText:
+		msgReq.Type = models.MessageTypeText
+		msgReq.Content = message.Content
+
+	case models.MessageTypeImage, models.MessageTypeVideo, models.MessageTypeAudio, models.MessageTypeDocument:
+		if message.MediaURL == "" {
+			return r.SendErrorEnvelope(fasthttp.StatusNotFound, "No media found for this message", nil, "")
+		}
+		data, errResp := a.readLocalMedia(r, message.MediaURL)
```
shridarpatil, line 100: **low** — `readLocalMedia` returns `nil` only on error, so a zero-byte stored media file passes the `data == nil` guard with `len(data) == 0`. `SendOutgoingMessage` uploads only `if mediaID == "" && len(req.MediaData) > 0` (`internal/handlers/messages.go:173`), so `mediaID` stays empty and `SendImageMessage`/`SendDocumentMessage` is called with an empty media id: the endpoint returns 200, the agent sees "Message forwarded", and the message fails at Meta with an opaque error.

Check `len(data) == 0` and send an explicit 404 — note you can't just swap the condition, since `errResp` is `nil` in that case and returning it would emit an empty 200.

internal/handlers/messages_forward.go
```diff
@@ -0,0 +1,167 @@
...
+
+	// Source message, org-guarded.
+	message, err := findByIDAndOrg[models.Message](a.DB, r, messageID, orgID, "Message")
+	if err != nil {
+		return nil
+	}
+
+	// The agent must be able to see the source conversation (same assignment
+	// scoping as the rest of the chat surface).
+	var sourceContact models.Contact
```
shridarpatil, line 57: **low** — the source-conversation check applies only `scopeAssignedContact`. `GetMessages` additionally restricts agents without `contacts:read` to messages from the current session when `AgentAssignment.CurrentConversationOnly` is set (`internal/handlers/contacts.go:289-302`). Forward skips that filter, so such an agent can forward a message from a *previous* session — content and media they can't view in the UI — to any contact they can reach, by passing its id directly.

`ServeMedia` has the same gap today, so this is a class issue rather than new, but forward widens it from "view" to "send to another contact".

frontend/src/views/chat/ChatView.vue
```diff
@@ -1181,6 +1182,67 @@ function replyToMessage(message: Message) {
...
+const isSearchingForward = ref(false)
+const forwardSendingId = ref<string | null>(null)
+let forwardSearchTimer: number | undefined
+
+const forwardableTypes = ['text', 'image', 'video', 'audio', 'document']
+function canForward(message: Message): boolean {
+  return forwardableTypes.includes(message.message_type)
+}
+
+function openForwardDialog(message: Message) {
```
shridarpatil, line 1202: **low** — `openForwardDialog` clears the query but not `forwardResults`, and `searchForwardContacts` filters against `contactsStore.currentContact?.id` at fetch time. Open the dialog in conversation A, close it, switch to conversation B, open it on a message: the stale list renders immediately and now contains B itself, so a fast click forwards to the same conversation and gets the backend's 400 "Cannot forward to the same conversation" as an error toast. Reset `forwardResults.value = []` here.

Same block, two related nits: the debounced search has no request sequencing, so a slow response for `"a"` can land after `"ab"` and overwrite it; and `forwardSearchTimer` (line 1193) is never cleared on dialog close or in `onUnmounted`, so a pending search fires after the dialog is gone.

frontend/src/views/chat/ChatView.vue
```diff
@@ -2678,6 +2760,51 @@ async function sendMediaMessage() {
...
+        <div class="py-4 space-y-3">
+          <div class="relative">
+            <Search class="absolute left-2.5 top-2.5 h-4 w-4 text-muted-foreground" />
+            <Input
+              v-model="forwardSearchQuery"
+              :placeholder="$t('chat.searchContacts') + '...'"
+              class="pl-9 h-9"
+            />
+          </div>
+          <ScrollArea class="max-h-[280px]">
```
shridarpatil: **low** — this `ScrollArea` won't scroll. `ScrollAreaRoot` gets `relative overflow-hidden` with no definite height and `ScrollAreaViewport` is `h-full`, so the percentage resolves against an auto-height containing block → `auto`; the viewport grows to full content height and the root just clips it. With the search capped at `limit: 20` (~36px per row ≈ 720px), roughly the last 13 contacts are unreachable.

Use a definite height like every other `ScrollArea` in the codebase (`h-[300px]`, `h-48`, `flex-1`). This mirrors the pre-existing assign-contact dialog at line 2653, but that list is client-filtered users and rarely overflows, whereas this one routinely will.

shridarpatil (commented):
Finding #1 (multipart header escaping in `UploadMedia`) is the one worth fixing before merge — forwarding is the first path that routes a remote contact's filename into that hand-built body. The other four are low severity.

Checked and deliberately **not** flagged: the missing `audit.LogAudit` call matches `SendMessage`/`SendMediaMessage` (message sends aren't audited anywhere in this codebase), and `readLocalMedia`'s path-traversal + symlink guards correctly mirror `ServeMedia`.

bambinounos:
Thanks for the review — all 5 findings addressed in 128764d (also cherry-picked to our production branch fbdcc0d):

**Medium (`pkg/whatsapp/client.go:324`)**: switched `UploadMedia` from hand-built multipart to `mime/multipart.Writer` via `CreatePart` with explicit header, plus CR/LF stripping on filename and mimeType. Filename escaping now mirrors `multipart.escapeQuotes` (backslash + double-quote). Also defaults empty mimeType to `application/octet-stream`.

**Low `messages_forward.go:100`**: now checks `len(data)==0` and returns 404 `Media file is empty` instead of forwarding zero bytes (which would hit `SendOutgoingMessage` with empty mediaID).

**Low `ChatView.vue:1200` (a) stale list**: `openForwardDialog` now resets `forwardResults`, and watch on `isForwardDialogOpen` clears query/results/message on close.
(b) stale search: added `forwardSearchSeq` monotonic counter; stale responses that land after a newer query are dropped.
(c) timer leak: `forwardSearchTimer` cleared on dialog close and in `onUnmounted`.

**Low `ChatView.vue:2781`**: `ScrollArea` changed from `max-h-[280px]` to `h-[280px]` so the 20-row result list actually scrolls (viewport `h-full` now has a definite height).

The `CurrentConversationOnly` gap you noted for forward vs `GetMessages` is intentional to keep this PR scoped — happy to follow up as a separate patch that also fixes `ServeMedia` for consistency if you prefer.

`go vet`/`go build ./...` clean; `vue-tsc` only shows the pre-existing `AccountDetailView.vue:167` error.

PR #510, opened 2026-07-18
feat(business-hours): optional CTA button on the out-of-hours message

bambinounos:
## Problem

The out-of-hours notice is text only. It can tell a customer the team is away, but it cannot hand them anywhere to go.

That is a gap for any business that keeps *something* running while the office is closed — a self-service bot, a help centre, an order form, a status page. Today the only way to point at it is pasting a raw URL into the message body, which looks like this on a phone:

> Estamos fuera de horario. Escríbele a nuestro asistente: https://t.me/example_bot

## Fix

Add `OutOfHoursButtonText` / `OutOfHoursButtonURL` to `BusinessHoursConfig`. When **both** are set, the notice goes out as a `cta_url` interactive message; otherwise it falls back to plain text exactly as before. Requiring both fields avoids emitting a malformed interactive message from a half-filled form.

WhatsApp allows a single CTA URL button per message and does not permit mixing it with reply buttons, so this is one label/URL pair rather than a list — matching the constraint rather than exposing an array the API cannot honour.

Almost all the plumbing was already in place (`SendCTAURLButton`, the `cta_url` case in the interactive dispatcher, `buildInteractiveData`, and the chat renderer that shows the button to agents). This wires the business-hours path into it and surfaces the two fields in the Business Hours settings tab.

Stacked on #509 — the diff here assumes the `sendOutOfHoursMessage` helper that PR introduces.

## Tests

`internal/handlers/chatbot_processor_test.go`:

- `TestSendOutOfHoursMessage_WithCTAButton` — with both fields set, the saved message is `interactive` with `interactive_data.type == "cta_url"` and the configured label and URL.
- `TestSendOutOfHoursMessage_ButtonNeedsBothFields` — a label with no URL falls back to a plain text message.

The existing out-of-hours tests are unchanged and still pass, pinning the text-only path.

## UI

Two inputs under the out-of-hours message in the Business Hours tab, with the label capped at 20 characters to match `SendCTAURLButton`'s truncation. Strings added to all five locales.


shridarpatil:
We now have a timing node in flow builder that can be used to handle out-of-hours.

PR #512, opened 2026-07-18
fix(frontend): restore session on public routes to redirect authenticated users away from login

ajasma6570:
## Description
Currently, when an authenticated user manually navigates to a public route like `/login` or `/register` (or hard refreshes the page), they are shown the login page instead of being correctly redirected to the dashboard.

This happens because the router navigation guard was only attempting to restore the user's session from `localStorage` if the route being navigated to explicitly required authentication.

## Changes Made
- Moved `authStore.restoreSession()` to the top of the `beforeEach` router guard in `frontend/src/router/index.ts`.
- It now runs unconditionally on the initial page load, ensuring the `isAuthenticated` state is correctly populated before deciding whether to show public pages.

## Steps to Reproduce the Original Bug
1. Log into the application successfully.
2. Manually type `/login` into the browser's address bar and hit enter.
3. Observe that you are kept on the login page instead of being redirected to your dashboard.

## Expected Behavior After Fix
Authenticated users who navigate to `/login` or `/register` are now immediately redirected to their first accessible route based on their role permissions.


shridarpatil:
## Review

Verified by applying the patch locally — typecheck and eslint are clean (the one pre-existing `vue-tsc` error in `AccountDetailView.vue:167` is unrelated to this PR).

### Overview

Single-file change in `frontend/src/router/index.ts`. Moves `authStore.restoreSession()` out of the `requiresAuth` branch and up to the top of the `beforeEach` guard, gated on `!authStore.isAuthenticated`. This populates the auth store before the `else` branch evaluates, so an authenticated user who lands on `/login` or `/register` (bookmark, address bar, hard refresh) is redirected to `getFirstAccessibleRoute()` instead of being shown the login form.

The diagnosis is right and the fix is the minimal one for it. The protected-route path is behaviorally identical — restore-then-bail collapses into the same outcome.

### Correctness

**Good:**

- The `!isAuthenticated` gate is the right call — it avoids re-parsing localStorage and clobbering `user.value` on every in-app navigation.
- No loop risk from a dead session. `restoreSession()` (`stores/auth.ts:82`) trusts localStorage without validating the cookie, so a stale `user` entry now bounces you off `/login` to the dashboard — but every authenticated route is nested under `AppLayout` (`router/index.ts:38-41`), whose `onMounted` calls `refreshUserData()` → `/me` (`AppLayout.vue:33-36`). That 401s, refresh fails, and the interceptor clears localStorage and hard-redirects to `/login` (`services/api.ts:141-144`). Self-heals reliably.
- Post-logout is fine: `clearAuth()` removes `user` from localStorage, so `restoreSession()` returns false.

**Worth flagging:**

- **Extra bounce + full page reload on a dead session (low, UX).** The self-heal above costs the user `/login` → dashboard → `window.location.href` reload → `/login`. Acceptable as-is. If you wanted it clean, the guard is already `async` — for `to.name === 'login' | 'register'` you could `await authStore.refreshUserData()` and only redirect on success. The trade-off is a `/me` round-trip on the login page for anyone with stale localStorage, so I'd ship as-is rather than pay that.
- **SSO error path can dead-end (low).** `/auth/sso/callback` is also `requiresAuth: false` (`router/index.ts:32-35`), so the store now gets populated from stale localStorage there too. On the success path this is harmless — `SSOCallbackView.vue:25` overwrites it via `setAuth()` from a fresh `/me`. But on the error path, the "Return to login" link (`SSOCallbackView.vue:72`) will now be bounced to the dashboard by the guard. Narrow (needs stale localStorage *and* no prior 401), and the dashboard's `/me` will kick them back — but it's new behavior on that route. To scope the change tightly, gate the restore on `to.name === 'login' || to.name === 'register'`.

### Conventions & style

- Matches the existing guard style; lint and typecheck are clean.
- Comment nit: *"Always try to restore session from localStorage on initial load"* — it's inside a `!isAuthenticated` check and fires on any navigation where the store is empty, not just initial load. Suggest: `// Hydrate the store from localStorage before any route decision — an authenticated user hitting /login must be redirected away.`
- The two consecutive `if (!authStore.isAuthenticated)` checks read a little redundant, but they mean different things (hydrate vs. bail). Fine to leave.
- Aside, out of scope: `ChatView.vue:426-427` does the same lazy `restoreSession()` dance. A single `restoreSession()` in `main.ts` before `app.mount()` would make the guard a pure state reader and delete both ad-hoc calls. Cleaner, but a separate PR — this one is correctly minimal.

### Test coverage

This is the main gap. `frontend/e2e/tests/auth/login.spec.ts` covers login, logout, validation, and unauthenticated→`/login`, but has nothing for the behavior being fixed. Per `frontend/e2e/ARCHITECTURE.md` (UI-driven by default), a short test belongs here:

```ts
test('should redirect authenticated user away from login page', async ({ page }) => {
  await loginPage.login(TEST_USERS.admin.email, TEST_USERS.admin.password)
  await loginPage.expectLoginSuccess()
  await page.goto('/login')
  await expect(page).not.toHaveURL(/\/login/)
})
```

Worth a `/register` case too, since the guard covers both.

### Security / performance

No concerns. No new data crosses a trust boundary; `restoreSession()` is synchronous localStorage access and is not a source of authorization — the backend still validates the httpOnly cookie on every request, and the store is only ever a UI hint. Performance impact is one extra `JSON.parse` on cold navigations to public routes.

### Verdict

Approve with minor changes — fix the comment wording and add the e2e test. Optionally narrow the restore to `login`/`register` to leave the SSO callback route untouched.

arunachalaservice:
I’ve pushed an update addressing the review comments.

The latest changes include:

- Narrowing restoreSession() to protected routes, /login, and /register, so the SSO callback route remains unaffected.
- Updating the router guard comment to better describe its intent.
- Adding Playwright E2E tests for authenticated redirects from both /login and /register.

I'd appreciate another look whenever you have a chance.

PR #515, opened 2026-07-21
feat(templates): template builder with live preview, validation, and payload preview

Rax498:
## Summary

Rebuilds the template create/edit page around a richer editor, keeping the app's
existing page structure (list at /templates, create/edit at /templates/new and /:id).

image

**What this adds vs main:**

- **Live WhatsApp-style preview** while creating and editing (previously only a plain
  preview after saving): WhatsApp text formatting (*bold*, _italic_, ~strike~, ```mono```),
  media headers, per-type button icons, and sample values substituted into variables.
- **Client-side validation of Meta's rules** with clear messages before submission:
  - character limits (header 60, body 1024, footer 60, button label 25)
  - variable rules: no mixing positional/named, sequential {{1}}, {{2}}…, no variables at
    body edges, sample values required
  - button rules: max 10 total, max 2 call-to-action (URL/phone), per-type caps, and
    same-type grouping — with reorder controls, since array order is what Meta receives
- **All 7 button types** the backend accepts: quick reply, URL (static and dynamic with
  example value), phone, copy code, flow, voice call, OTP.
- **AUTHENTICATION templates** fully supported in the editor: OTP delivery type
  (copy code / one-tap / zero-tap + supported apps), security recommendation,
  code expiration.
- **Payload inspection before sending**: a JSON dialog showing both the exact API
  request body and the exact Meta payload (built by the backend — see below).
- **Media headers preview locally** (object URL); the file is uploaded to Meta only on
  save, so abandoned drafts no longer leave orphaned upload handles.
- **Publish a draft directly from the templates list.**

## Backend changes (3 files, no behaviour change)

- `cmd/whatomate/main.go` — 1 line: registers `POST /api/templates/preview`
- `internal/handlers/templates.go` — new `PreviewTemplate` handler. Read-only: builds
  the payload in memory and returns it. No DB write, no Meta call.
- `pkg/whatsapp/template.go` — moves the payload-wrapping block out of
  `SubmitTemplate` into `BuildSubmissionPayload`, now called by both `SubmitTemplate`
  and the preview. The component builders are untouched; publishing sends the same
  payload as before.

## e2e spec updates

All 36 template specs pass. Four expectations were updated minimally because this PR
replaces UI they encoded (the collapsible Details card and the Preview button + modal
are superseded by the always-visible editor with a live sidebar preview); test intent
is preserved. Also fixed a self-colliding locator in the media-header spec that matched
the spec's own seeded account name.

## Known limitation

Media headers are previewed live while creating a template (local object URL).
For already-saved templates the preview shows a placeholder instead of the real
image: whatomate deliberately stores only Meta's upload handle, which is not
downloadable, and keeps no copy of the media. A possible follow-up is rendering
the temporary media URL Meta returns during template sync.


frontend/src/views/settings/TemplateDetailView.vue
```diff
@@ -242,132 +195,89 @@ const hasTooManyHeaderVariables = computed(() => {
...
+    if (unique[i] !== i + 1) return { expected: i + 1, found: unique[i] }
+  }
+  return null
+})
 
-function setSampleValueForVar(component: string, index: number, value: string) {
-  const existing = form.value.sample_values.findIndex(
-    (s: any) => s.component === component && s.index === index
+// Meta rejects a template whose variables have no example values.
+const missingSamples = computed(() =>
```
shridarpatil, line 212: **A named variable used twice can never be saved.**

`extractParamNames()` in `TemplateEditor.vue` dedupes (it keeps a `seen` set), so the editor renders **one** input per unique name. `allVariables` here counts every *occurrence* and indexes by occurrence position. The comment on line 153 says the two must agree — they diverge exactly on duplicates.

Body `Hi {{name}}, your order for {{item}} ships today. Thanks {{name}} for shopping.`:

```
editor inputs : [ 'name', 'item' ]    -> writes samples index 1, 2
allVariables  : name@1, item@2, name@3
missingSamples: [ '{{name}}' ]        -> save() bails at line 506
```

The toast reads "Add sample values for: `{{name}}`" and there is no input that can satisfy it — the template becomes unsavable.

Same for a TEXT header `Hi {{name}} - welcome {{name}}`, which is precisely the case the comment on line 192 claims passes: `hasTooManyHeaderVariables` dedupes with a `Set`, `missingSamples` does not.

Fix: dedupe `allVariables` by `(component, name)` so it matches `sampleIndexFor()`, or drop the occurrence-position indexing for named params.

frontend/src/views/settings/TemplateEditor.vue
```diff
@@ -0,0 +1,1612 @@
...
+  return slot >= 0 ? slot + 1 : 1;
+}
+
+function findSample(component: string, paramName: string) {
+  const samples = state.value.sample_values || [];
+  const index = sampleIndexFor(component, paramName);
+  return samples.find(
+    (s: any) =>
+      s.component === component &&
+      (s.param_name === paramName || s.index === index),
```
shridarpatil, line 445: **Deleting a non-final named variable silently submits another variable's sample.**

The `s.index === index` fallback fires for named params too, so a sample can be matched by slot even when its `param_name` is something else. Remove `{{name}}` from `Hi {{name}}, we ship to {{city}} tomorrow.`:

```
samples before      : name@1="Ada", city@2="Berlin"
after prune watcher : name@1="Ada"            (TemplateDetailView.vue:261)
findSample('body','city') -> {"param_name":"name","value":"Ada"}
missingSamples      : []                      -> save proceeds
```

So the `{{city}}` input displays `Ada` with no warning, and the payload keeps `param_name: "name"`. `extractNamedExamplesForComponent` (`pkg/whatsapp/template.go:545`) keys named examples off `param_name`, so Meta gets `body_text_named_params: [{param_name:"name", ...}]` for a body containing `{{city}}` and rejects the template with an opaque error.

`TemplatePreview.vue:86-96` already gets this right — index only when the token is numeric, `param_name` otherwise. Worth reusing that rule here.

frontend/src/views/settings/TemplateEditor.vue
```diff
@@ -0,0 +1,1612 @@
...
+        <!-- Meta forces header_type NONE on AUTHENTICATION templates -->
+        <div
+          v-if="!isAuthentication"
+          class="space-y-4 p-4 border rounded-lg bg-muted/10"
+        >
+          <div class="grid grid-cols-2 gap-4">
+            <div class="space-y-1.5">
+              <div class="flex items-center h-7">
+                <Label class="text-xs font-bold text-muted-foreground"
+                  >Header Type</Label
```
shridarpatil, line 791: **i18n regression.** These labels were `$t()` on `main` (e.g. `main:767` had `$t('templates.headerType', 'Header Type')`); this file hardcodes ~33 user-visible strings — Header Type, Header Text, Message Body, Footer Text, Action Buttons, Sample Values for Variables, Code Delivery Method, Supported Apps, the OTP explanations, "Before you submit", the Flow/Screen selects and their placeholders. All five locales lose them.

frontend/src/views/settings/TemplateEditor.vue
```diff
@@ -0,0 +1,1612 @@
...
+
+function clearMediaFile() {
+  mediaFileName.value = "";
+  state.value.header_content = "";
+  emit("update:mediaFile", null);
+}
+
+onMounted(() => {
+  if (state.value?.buttons) {
+    state.value.buttons = state.value.buttons.map((b: any) => ({
```
shridarpatil, line 643: **Button `v-for` keys are lost after every save.**

Ids are assigned in `onMounted` only. `save()` -> `loadTemplate()` -> `syncForm()` rebuilds `form.buttons` from the server response, which carries no ids (`cleanButton` strips them on the way out). The props watcher on line 117 then sees a `JSON.stringify` difference and replaces `state` wholesale, so `:key="btn.id"` on line 1326 is `undefined` for every button — duplicate-key warnings, and the keyed diff degenerates to positional.

Buttons the parent creates later have the same problem: the category watcher at `TemplateDetailView.vue:358` pushes an OTP button with no id, and `onMounted` has already run.

Normalizing ids inside the props watcher (or keying on index) fixes both.

frontend/e2e/tests/settings/template-sample-values.spec.ts
```diff
@@ -136,14 +136,9 @@ test.describe('Template Preview with Sample Values', () => {
...
-
-    const dialog = page.locator('[role="alertdialog"]')
-    await expect(dialog).toBeVisible({ timeout: 5000 })
-    await expect(dialog.getByText('Template Preview')).toBeVisible()
-
-    await dialog.getByRole('button', { name: /Close/i }).click()
+    // The preview renders live in the sidebar; the body's variable substitution
+    // happens inline, so the template text is visible without opening anything.
+    await expect(page.getByText('Live Preview')).toBeVisible({ timeout: 10000 })
+    await expect(page.getByText(/welcome/).first()).toBeVisible()
```
shridarpatil, line 142: This test is still named "should show preview with sample values replacing variables", but the template it seeds has **no** `sample_values` at all, so it can't assert substitution — it now only checks that the sidebar heading and the body text render.

The old version was equally vacuous (it just opened the dialog and closed it), so no coverage is lost. But the headline feature of the PR — samples substituted into the live preview — ends up with no assertion anywhere, and neither do `validateButtonCombination`, the Meta-payload dialog, or any of the new client-side validations. Seeding `sample_values` and asserting the green pill contains the value is cheap, and would have caught the two indexing bugs I flagged in `TemplateEditor.vue` / `TemplateDetailView.vue`.

frontend/src/views/settings/TemplateEditor.vue
```diff
@@ -0,0 +1,1612 @@
...
+
+          <!-- ONE_TAP / ZERO_TAP: the app that receives the autofilled code -->
+          <div
+            v-if="authOtpType === 'ONE_TAP' || authOtpType === 'ZERO_TAP'"
+            class="space-y-3 rounded-lg border p-3"
+          >
+            <div v-if="authOtpType === 'ONE_TAP'" class="space-y-1">
+              <Label class="text-xs">Autofill Text</Label>
+              <Input
+                v-model="otpButton().autofill_text"
```
shridarpatil, line 1038: Only unguarded `otpButton()` call in the file — every other one uses `otpButton()?.`. Reachable-safe today because `authOtpType` returns `COPY_CODE` when no OTP button exists, but it's one refactor away from a render-time TypeError.

Worth knowing while you're here: `buildAuthComponents` (`pkg/whatsapp/template.go`) reads flat `package_name` / `signature_hash` off the button and never looks at `supported_apps`, and ignores `autofill_text` entirely. So the ONE_TAP/ZERO_TAP fields this editor now validates as *required* (`TemplateDetailView.vue:529-535`) never reach Meta. Pre-existing on `main`, but it undercuts the "AUTHENTICATION templates fully supported" claim.

frontend/src/views/settings/TemplateEditor.vue
```diff
@@ -0,0 +1,1612 @@
...
+  return /^\+?[0-9]{7,15}$/.test(value.trim());
+}
+
+// Meta treats a URL button as dynamic when the url ends in {{1}}, so the url is
+// the source of truth. The example suffix goes on the button as `example`, which
+// is where the backend reads it from.
+const URL_VAR = "{{1}}";
+
+function isDynamicUrl(btn: any) {
+  return String(btn.url || "").includes(URL_VAR);
```
shridarpatil, line 82: Only `{{1}}` is recognised as dynamic. A URL button synced from Meta that uses a named param (`https://x/{{order}}`) shows as **Static**, so no example is collected and `firstButtonError` doesn't ask for one — then the backend's URL branch sees `strings.Contains(btnURL, "{{")` and sends the URL with no `example`, which Meta rejects.

frontend/src/views/settings/TemplateEditor.vue
```diff
@@ -0,0 +1,1612 @@
...
+function saveCursorPosition(event: Event) {
+  const el = event.target as HTMLTextAreaElement;
+  if (el && typeof el.selectionStart === "number") {
+    savedSelectionStart.value = el.selectionStart;
+    savedSelectionEnd.value = el.selectionEnd ?? el.selectionStart;
+  }
+}
+
+function insertAtCursor(textToInsert: string, cursorOffset: number = 0) {
+  const startPos = savedSelectionStart.value;
```
shridarpatil, line 267: `savedSelectionStart` starts at `0` and is only updated on `blur`/`keyup`/`mouseup`, so if the textarea was never focused, "Add Variable" and the format buttons insert at the very beginning of the body rather than the end.

frontend/src/views/settings/TemplateEditor.vue
```diff
@@ -0,0 +1,1612 @@
...
+    input.value = "";
+    return;
+  }
+
+  mediaFileName.value = file.name;
+  state.value.header_content = "";
+  emit("update:mediaFile", file);
+}
+
+function clearMediaFile() {
```
shridarpatil, line 635: `mediaFileName` is never cleared when the parent nulls `pendingMediaFile` after a successful upload in `save()`, so the filename chip persists instead of switching to "Sample already uploaded. Choose a file to replace it."

frontend/src/views/settings/TemplatesView.vue
```diff
@@ -251,6 +258,24 @@ async function confirmDeleteTemplate() {
   }
 }
 
+// A DRAFT template lives only in our database until it is submitted to Meta.
+function isDraft(template: Template) {
+  return !template.meta_template_id || template.status?.toUpperCase() === 'DRAFT'
```
shridarpatil, line 263: `isDraft()` offers Publish for `!meta_template_id || DRAFT`, but the detail page's `canPublish` uses `DRAFT || REJECTED`. A REJECTED template can be republished from the detail page but not from the list — worth making the two agree.

Also note this action (and Edit/Delete in the new overflow menu) isn't gated on `templates:write`/`delete` the way the detail page gates its buttons. Delete was already ungated here on `main`, so no regression, just carrying it forward to a new action.

internal/handlers/templates.go
```diff
@@ -430,6 +430,46 @@ func (a *App) SubmitTemplate(r *fastglue.Request) error {
...
+	if _, err := a.getOrgID(r); err != nil {
+		return r.SendErrorEnvelope(fasthttp.StatusUnauthorized, "Unauthorized", nil, "")
+	}
+
+	var req PreviewTemplateRequest
+	if err := a.decodeRequest(r, &req); err != nil {
+		return nil
+	}
+
+	payload, err := a.WhatsApp.BuildSubmissionPayload(&whatsapp.TemplateSubmission{
```
shridarpatil, line 452: Checked the fidelity claim and it holds: `submitTemplateToMeta` also leaves `ParameterFormat` empty, so both paths derive `parameter_format` purely from `hasNamedParams(body)`, and `CreateTemplate` normalizes `Name`/`Category`/`HeaderType` exactly as this does. The preview really is the same payload.

The extraction into `BuildSubmissionPayload` is faithful too — `template.MetaTemplateID != ""` is the same condition the old inline `isUpdate` used. No permission check here matches the rest of `templates.go`, which has none either.

shridarpatil (commented):
Read through the whole diff. The `BuildSubmissionPayload` extraction is a faithful refactor, and `TemplatePreview` escapes then DOMPurify-sanitizes with a function replacer so sample values can't inject `$&`/`$1` — no XSS there. The `DropdownMenu` `open: undefined` fix also unbreaks ChatView's contact-options menu, the only other consumer, which was dead on `main` for the same reason.

Two data-correctness bugs in the new sample-value indexing, both reproduced by replaying the actual functions:

1. A named variable used twice makes the template **unsavable** — the editor dedupes variable names, `missingSamples` doesn't, so it demands a sample for an input that isn't rendered. This is the case the comment on `TemplateDetailView.vue:192` claims passes.
2. Deleting a non-final named variable makes `findSample`'s index fallback hand `{{city}}` the value that belonged to `{{name}}`, and ships `param_name: "name"` to Meta.

Plus an i18n regression (~33 labels that were `$t()` on `main` are now hardcoded, 18 new keys never added to `en.json`, and the Crowdin-managed locale files hand-edited), and button `v-for` keys going `undefined` after every save. Details inline.

Verification: `eslint` clean on all changed frontend files; `vue-tsc --noEmit` shows only one pre-existing error in `AccountDetailView.vue` (untouched here). Go `build`/`vet` couldn't run in my sandbox (module proxy unreachable), so `templates.go` / `template.go` were reviewed by reading only.

PR #519, opened 2026-07-25
feat(chat): in-app media viewer (lightbox) for images, video and PDF

bambinounos:
## What

Adds an in-app media viewer (lightbox) for chat attachments. Today, clicking an image opens a new browser tab (`window.open`) and documents force a download — there is no in-app viewer.

The viewer is built entirely on the existing `ui/dialog` primitives, so ESC-to-close, backdrop-click and focus-trap come for free. Features:

- Images / stickers at full size with **zoom** (wheel / buttons / double-click) and **drag-to-pan**.
- Native video playback.
- **Inline PDF** rendering.
- **Prev/next navigation** across all media in the conversation, with arrow keys.
- Toolbar: filename, download, open-in-tab.

## PDF: working around `X-Frame-Options: DENY`

The media endpoint (`/api/media/{id}`) serves PDFs inline with the correct MIME type, but `SecurityHeaders()` sets a global `X-Frame-Options: DENY`, which blocks a direct `<iframe src="/api/media/{id}">` even same-origin.

Rather than weaken that header, the viewer fetches the bytes with `fetch(url, { credentials: 'same-origin' })` (the `whm_access` cookie rides along) and frames an **object-URL blob**. `X-Frame-Options` does not apply to `fetch`, and the `blob:` response carries no such header, so the browser's native PDF viewer renders it. A download fallback covers WebViews that won't render inline. **No backend / security change.**

Images and video load straight from `src` (the cookie authenticates them; XFO doesn't apply to `` / `<video>`).

## Scope

- New component: `frontend/src/components/chat/MediaViewerDialog.vue`.
- `ChatView.vue`: `openMediaPreview` now opens the viewer instead of a new tab; a `viewableMedia` computed backs the gallery; the two document bubbles become clickable (download stays available inside the viewer).
- i18n keys `chat.mediaViewer.*` (en / es).
- **Frontend only, no Go changes.**

`npm run build` passes; `npm run typecheck` is clean apart from a pre-existing, unrelated error in `AccountDetailView.vue` (`business_calling_enabled`).

### DIFF
--- a/frontend/src/components/chat/MediaViewerDialog.vue
+++ b/frontend/src/components/chat/MediaViewerDialog.vue
@@ -1,0 +1,281 @@
+<script setup lang="ts">
+// In-app media viewer (lightbox) for chat attachments: images, stickers, video
+// and PDF/documents. Replaces the old "open in a new browser tab" behaviour.
+//
+// Built on the existing ui/dialog primitives, so teleport-to-body, the dark
+// backdrop, ESC-to-close and focus-trap all come for free. Images and video
+// load straight from the auth-cookie-gated /api/media/{id} endpoint. PDFs are
+// the tricky case: the backend sets a global `X-Frame-Options: DENY`, so a
+// direct <iframe src="/api/media/{id}"> is blocked even same-origin. We work
+// around it WITHOUT touching the backend by fetch()ing the bytes (the cookie
+// rides along) and framing an object-URL blob, which carries no such header.
+import { ref, computed, watch, onBeforeUnmount } from 'vue'
+import { onKeyStroke } from '@vueuse/core'
+import { Dialog, DialogContent } from '@/components/ui/dialog'
+import IconButton from '@/components/shared/IconButton.vue'
+import { Button } from '@/components/ui/button'
+import {
+  Download,
+  ExternalLink,
+  ZoomIn,
+  ZoomOut,
+  ChevronLeft,
+  ChevronRight,
+  FileText,
+  Loader2,
+} from 'lucide-vue-next'
+import type { Message } from '@/stores/contacts'
+
+const props = defineProps<{ items: Message[] }>()
+const open = defineModel<boolean>('open', { default: false })
+const index = defineModel<number>('index', { default: 0 })
+
+const current = computed<Message | undefined>(() => props.items[index.value])
+const total = computed(() => props.items.length)
+const hasPrev = computed(() => index.value > 0)
+const hasNext = computed(() => index.value < total.value - 1)
+
+// Same URL the chat bubbles use: relative, same-origin, auth-cookie gated.
+function mediaUrl(m: Message): string {
+  if (!m.media_url) return ''
+  const basePath = ((window as any).__BASE_PATH__ ?? '').replace(/\/$/, '')
+  return `${basePath}/api/media/${m.id}`
+}
+
+type Kind = 'image' | 'video' | 'pdf' | 'file'
+function mediaKind(m: Message): Kind {
+  const mime = m.media_mime_type || ''
+  const name = (m.media_filename || '').toLowerCase()
+  if (m.message_type === 'image' || m.message_type === 'sticker' || mime.startsWith('image/')) return 'image'
+  if (m.message_type === 'video' || mime.startsWith('video/')) return 'video'
+  if (mime.includes('pdf') || name.endsWith('.pdf')) return 'pdf'
+  return 'file'
+}
+const currentKind = computed<Kind>(() => (current.value ? mediaKind(current.value) : 'file'))
+const currentUrl = computed(() => (current.value ? mediaUrl(current.value) : ''))
+const filename = computed(() => current.value?.media_filename || '')
+
+// --- Navigation ---------------------------------------------------------
+function prev() { if (hasPrev.value) index.value -= 1 }
+function next() { if (hasNext.value) index.value += 1 }
+onKeyStroke('ArrowLeft', () => { if (open.value) prev() })
+onKeyStroke('ArrowRight', () => { if (open.value) next() })
+
+function openInTab() { if (currentUrl.value) window.open(currentUrl.value, '_blank') }
+
+// --- Image zoom / pan ---------------------------------------------------
+const scale = ref(1)
+const tx = ref(0)
+const ty = ref(0)
+const dragging = ref(false)
+let startX = 0, startY = 0, startTx = 0, startTy = 0
+const MIN = 1, MAX = 5, STEP = 0.25
+
+function resetZoom() { scale.value = 1; tx.value = 0; ty.value = 0 }
+function clamp(v: number) { return Math.min(MAX, Math.max(MIN, +v.toFixed(2))) }
+function zoomIn() { scale.value = clamp(scale.value + STEP) }
+function zoomOut() { scale.value = clamp(scale.value - STEP); if (scale.value === 1) { tx.value = 0; ty.value = 0 } }
+function onWheel(e: WheelEvent) {
+  e.preventDefault()
+  scale.value = clamp(scale.value + (e.deltaY < 0 ? STEP : -STEP))
+  if (scale.value === 1) { tx.value = 0; ty.value = 0 }
+}
+function onPointerDown(e: PointerEvent) {
+  if (scale.value <= 1) return
+  dragging.value = true
+  startX = e.clientX; startY = e.clientY; startTx = tx.value; startTy = ty.value
+  ;(e.target as HTMLElement).setPointerCapture?.(e.pointerId)
+}
+function onPointerMove(e: PointerEvent) {
+  if (!dragging.value) return
+  tx.value = startTx + (e.clientX - startX)
+  ty.value = startTy + (e.clientY - startY)
+}
+function onPointerUp() { dragging.value = false }
+function toggleZoom() { scale.value > 1 ? resetZoom() : (scale.value = 2) }
+
+const imageStyle = computed(() => ({
+  transform: `translate(${tx.value}px, ${ty.value}px) scale(${scale.value})`,
+  cursor: scale.value > 1 ? (dragging.value ? 'grabbing' : 'grab') : 'zoom-in',
+  transition: dragging.value ? 'none' : 'transform 0.15s ease',
+}))
+
+// --- PDF blob (X-Frame-Options: DENY workaround) ------------------------
+const pdfBlobUrl = ref<string | null>(null)
+const pdfLoading = ref(false)
+const pdfError = ref(false)
+let pdfReqId = 0
+
+function clearPdf() {
+  if (pdfBlobUrl.value) { URL.revokeObjectURL(pdfBlobUrl.value); pdfBlobUrl.value = null }
+  pdfLoading.value = false
+  pdfError.value = false
+}
+async function loadPdf(m: Message) {
+  const my = ++pdfReqId
+  clearPdf()
+  pdfLoading.value = true
+  try {
+    const res = await fetch(mediaUrl(m), { credentials: 'same-origin' })
+    if (!res.ok) throw new Error(`HTTP ${res.status}`)
+    const blob = await res.blob()
+    if (my !== pdfReqId) return // superseded by a newer navigation
+    pdfBlobUrl.value = URL.createObjectURL(blob)
+  } catch {
+    if (my === pdfReqId) pdfError.value = true
+  } finally {
+    if (my === pdfReqId) pdfLoading.value = false
+  }
+}
+
+// React to open / navigation: reset zoom and (re)load PDFs on demand.
+watch(
+  [open, index, () => props.items],
+  () => {
+    resetZoom()
+    const m = current.value
+    if (open.value && m && mediaKind(m) === 'pdf') loadPdf(m)
+    else clearPdf()
+  },
+  { immediate: true },
+)
+onBeforeUnmount(clearPdf)
+
+function closeOnBackdrop() { open.value = false }
+</script>
+
+<template>
+  <Dialog v-model:open="open">
+    <DialogContent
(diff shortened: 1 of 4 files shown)


frontend/src/components/chat/MediaViewerDialog.vue
```diff
@@ -0,0 +1,281 @@
...
+  } catch {
+    if (my === pdfReqId) pdfError.value = true
+  } finally {
+    if (my === pdfReqId) pdfLoading.value = false
+  }
+}
+
+// React to open / navigation: reset zoom and (re)load PDFs on demand.
+watch(
+  [open, index, () => props.items],
```
shridarpatil: **Zoom resets and the PDF re-fetches on every incoming message / delivery receipt.**

`props.items` is the `viewableMedia` computed, whose `.filter()` allocates a **new array on every recompute**. `contactsStore.messages` is mutated by `push()` on inbound WS messages (`stores/contacts.ts:315`) and by index assignment on status updates (`stores/contacts.ts:322`) — both invalidate the computed, so the watcher fires on a new array identity even though nothing about the current item changed.

With the viewer open, a `sent → delivered → read` receipt in that conversation therefore:
- snaps zoom/pan back to 1×, and
- for a PDF, revokes the blob, flashes the spinner, and re-fetches the whole file with the iframe reset to page 1.

Watching identity instead of the array fixes it, and also covers the index-shift from `loadOlderMessages` prepending (`ChatView.vue:217`):

```suggestion
  [open, () => current.value?.id],
```

frontend/src/components/chat/MediaViewerDialog.vue
```diff
@@ -0,0 +1,281 @@
...
+)
+onBeforeUnmount(clearPdf)
+
+function closeOnBackdrop() { open.value = false }
+</script>
+
+<template>
+  <Dialog v-model:open="open">
+    <DialogContent
+      class="max-w-none w-screen h-screen p-0 gap-0 border-0 rounded-none sm:rounded-none bg-black/95 shadow-none ring-0 flex flex-col"
```
shridarpatil: **The built-in close ✕ is invisible in light theme.**

`DialogContent` renders its `DialogClose` with no text colour (`ui/dialog/DialogContent.vue:38`) — it normally sits on `bg-background`. Here the background is overridden to `bg-black/95`, but `text-white` is set only on the toolbar `<div>`, which is a *sibling* of the close button. The portal target is `body`, which is `text-foreground` = `hsl(0 0% 9%)` in light mode (`assets/index.css:48,112`), so the ✕ renders near-black at `opacity-70` on a near-black backdrop.

ESC and backdrop-click still work, so it's cosmetic — but easy to fix:

```suggestion
      class="max-w-none w-screen h-screen p-0 gap-0 border-0 rounded-none sm:rounded-none bg-black/95 text-white shadow-none ring-0 flex flex-col"
```

frontend/src/components/chat/MediaViewerDialog.vue
```diff
@@ -0,0 +1,281 @@
...
+    >
+      <!-- Toolbar (pr-14 leaves room for DialogContent's built-in close X) -->
+      <div class="flex items-center gap-1 shrink-0 px-3 py-2 pr-14 text-white">
+        <span class="text-sm truncate max-w-[40vw]" :title="filename">{{ filename }}</span>
```
shridarpatil: **No `DialogTitle`.** Every other feature dialog in the repo has one — the only `DialogContent` users without a title are the `ui/` primitives themselves. reka-ui logs a dev warning and screen readers get an unlabeled modal.

Wrapping this filename span in a `DialogTitle` (it's already the accessible name of the dialog) is the cheapest fix; add a `VisuallyHidden` `DialogDescription` too if you want the `aria-describedby` warning gone as well.

frontend/src/components/chat/MediaViewerDialog.vue
```diff
@@ -0,0 +1,281 @@
...
+const pdfLoading = ref(false)
+const pdfError = ref(false)
+let pdfReqId = 0
+
+function clearPdf() {
+  if (pdfBlobUrl.value) { URL.revokeObjectURL(pdfBlobUrl.value); pdfBlobUrl.value = null }
+  pdfLoading.value = false
+  pdfError.value = false
+}
+async function loadPdf(m: Message) {
```
shridarpatil, line 121: Non-blocking, but worth a comment here so the next person doesn't remove the guardrail unknowingly.

`blob:` URLs inherit the **creating document's origin**, and this app ships no CSP at all (`internal/middleware/middleware.go:98-107` sets only XFO / nosniff / Referrer-Policy). So the only thing keeping this iframe from being a same-origin XSS sink is that `ServeMedia` maps the on-disk extension to a whitelisted `Content-Type`, defaulting to `application/octet-stream` (`internal/handlers/media.go:210-247`), and never trusts `media_mime_type` from the DB. (`saveMediaLocally` will take the extension from the client filename for unknown MIMEs — `internal/handlers/contacts.go:901-910` — but an unrecognised extension still serves as octet-stream.)

A one-line comment noting "safe only while `ServeMedia` whitelists Content-Type" would make that dependency explicit.

frontend/src/components/chat/MediaViewerDialog.vue
```diff
@@ -0,0 +1,281 @@
...
+          controls
+          class="max-w-full max-h-full"
+        />
+
+        <!-- PDF -->
+        <template v-else-if="current && currentKind === 'pdf'">
+          <div v-if="pdfLoading" class="flex flex-col items-center gap-2 text-white/80">
+            <Loader2 class="h-8 w-8 animate-spin" />
+          </div>
+          <iframe
```
shridarpatil, line 274: Nit: no failure path once the fetch succeeds. If the blob's type isn't `application/pdf` — e.g. an inbound PDF whose MIME didn't match `getExtensionFromMimeType`, saved as `.bin` and served as octet-stream — `mediaKind` still returns `pdf` from the `.pdf` filename, so you get a blank white frame rather than the `pdfError` UI. Recoverable via the toolbar, just not obvious. Checking `blob.type` before setting `pdfBlobUrl` would route those to the download fallback.

frontend/src/views/chat/ChatView.vue
```diff
@@ -2120,16 +2138,16 @@ async function sendMediaMessage() {
                 </div>
                 <!-- Document message -->
                 <div v-else-if="message.message_type === 'document' && message.media_url" class="mb-2">
-                  <a
-                    :href="getMediaUrl(message)"
-                    :download="message.media_filename || 'document'"
-                    class="flex items-center gap-2 px-3 py-2 bg-background/50 rounded-lg hover:bg-background/80 transition-colors"
+                  <button
```
shridarpatil, line 2172: Nit / product call rather than a defect: document bubbles go from one-click download to click → modal → click Download, and `docx`/`xlsx` have no inline viewer to justify the detour. Only PDFs really gain from routing through the lightbox. Worth considering keeping the anchor for non-PDF documents, or at least confirming the extra click is intended for the common case.

shridarpatil (commented):
Nice piece of work — the fetch→blob workaround is the right call (XFO `DENY` blocks same-origin framing too, and `<object>`/`<embed>` are equally blocked), and building on `ui/dialog` keeps it small.

Verified locally on this branch: `npm run typecheck` shows only the pre-existing `AccountDetailView.vue` error you called out, and `eslint` on both changed files is clean (the `no-empty` hits in `ChatView.vue` are pre-existing, untouched lines).

One behavioural bug I'd like fixed before merge (zoom/PDF resetting on every delivery receipt), two one-line polish items, plus some notes. Details inline.

On the security angle — I checked, and the blob-in-iframe is safe **as the backend stands today**: `ServeMedia` derives `Content-Type` from the on-disk extension against a fixed whitelist and never echoes the client-supplied MIME, so an HTML payload named `invoice.pdf` frames as `application/octet-stream`, not as scriptable same-origin content. Worth knowing that this is load-bearing, since the app ships no CSP — see the inline note on `loadPdf`.

frontend/src/components/chat/MediaViewerDialog.vue
bambinounos: Confirmed and fixed in f816b67 — thank you, this was a real bug and I'd been staring past it.

I reproduced it exactly as you described: with a PDF open, the `sent → delivered → read` sequence on any message in the conversation fires the watcher three times, because `viewableMedia`'s `.filter()` hands back a fresh array each recompute even when nothing in it changed.

Took your suggestion verbatim — `[open, () => current.value?.id]`. Watching identity rather than position also fixes the `loadOlderMessages` prepend case you flagged, where the index shifts under a stationary item and the old watcher would have re-fetched. Status-update mutations that replace the object at an index now no-op, since the id is unchanged.

frontend/src/components/chat/MediaViewerDialog.vue
bambinounos: Fixed in f816b67 — applied your suggestion. Reproduced in light theme first; the ✕ was there the whole time, just `hsl(0 0% 9%)` at `opacity-70` on `bg-black/95`.

Moving `text-white` to `DialogContent` is the right level: everything inside inherits it, and the toolbar's own `text-white` is now redundant but harmless, so I left it as documentation of intent. Added a short comment above `DialogContent` noting *why* the colour lives there, since it looks removable otherwise.

frontend/src/components/chat/MediaViewerDialog.vue
bambinounos: Added in f816b67. The filename span is now a `DialogTitle` with `text-sm font-normal` (twMerge drops the primitive's `text-lg font-semibold`, so it looks identical).

One wrinkle: inbound images usually carry no `media_filename`, so the title would have been empty — an unlabeled modal either way. It falls back to a generic `chat.mediaViewer.title` ("Attachment"), which also fills the previously blank toolbar slot for images.

Also added the `sr-only` `DialogDescription` to clear the `aria-describedby` warning. I used `class="sr-only"` rather than reka-ui's `VisuallyHidden` to match the existing idiom in the repo (`ChatbotFlowBuilderView.vue:942`) and avoid a new import pattern.

frontend/src/components/chat/MediaViewerDialog.vue
bambinounos: Added in f816b67 — agreed this needed to be written down, since nothing in this file hints that the guardrail lives in `media.go`.

The comment names the dependency explicitly (whitelisted `Content-Type` from the on-disk extension, `application/octet-stream` fallback, `media_mime_type` never echoed) and is marked load-bearing. I also enforced it client-side per your other note, so the invariant is now checked on both sides rather than only assumed here.

frontend/src/components/chat/MediaViewerDialog.vue
bambinounos: Good catch — fixed in f816b67:

```ts
const blob = await res.blob()
if (my !== pdfReqId) return
if (!blob.type.includes('pdf')) return
pdfBlobUrl.value = URL.createObjectURL(blob)
```

A `.pdf` saved as `.bin` and served as octet-stream now falls through to the download card instead of a blank frame. Returning without setting `pdfError` was deliberate: the file isn't broken, we just can't render it, so the card shows the filename with Download / Open-in-tab rather than an error message. `pdfError` stays reserved for a genuinely failed fetch.

This doubles as the client-side half of the security invariant in the other thread — only bytes the server itself typed as PDF ever reach an iframe.

frontend/src/views/chat/ChatView.vue
bambinounos: Fair call, and you're right that it was an unintended regression rather than a deliberate trade — I routed all documents through the viewer without separating the case that actually benefits.

Fixed in f816b67 with an `isPreviewableDocument()` check (mime contains `pdf`, or filename ends in `.pdf`): PDFs open in the lightbox, everything else keeps the original `<a :download>` and its one click. Applied to both document bubbles — the standalone one and the template-header variant, which had the same problem.

Non-PDF documents stay in `viewableMedia`, so paging through the gallery still surfaces them with a download card. That keeps the arrows continuous across the conversation's attachments and the `file` branch of the viewer still has a purpose.

bambinounos:
Thanks for the thorough review — and for digging into `ServeMedia` to check the framing invariant yourself. All six points are addressed in f816b67.

**The blocking bug.** Your diagnosis was exactly right: `viewableMedia` is a computed `.filter()`, so it reallocates on every recompute, and the WS `push` / status-update assignment on `contactsStore.messages` invalidate it. With a PDF open, a single `sent → delivered → read` sequence fired the watcher three times, each one revoking the blob and re-fetching the file. Now watching `[open, () => current.value?.id]`, which no-ops on those recomputes and also handles the `loadOlderMessages` index shift.

**Polish.**
- `text-white` moved to `DialogContent` so the built-in close ✕ inherits it (reproduced the near-black-on-black in light theme first).
- `DialogTitle` on the filename, with a generic fallback since inbound images carry no `media_filename`, plus an `sr-only` `DialogDescription`. Used `class="sr-only"` rather than `VisuallyHidden` to match `ChatbotFlowBuilderView.vue:942`.

**Notes.**
- Added the load-bearing comment on `loadPdf` naming the `ServeMedia` Content-Type whitelist as the thing that keeps the blob-in-iframe safe, and enforced it client-side with a `blob.type` check — which also fixes your blank-frame case: a `.pdf` served as octet-stream now falls through to the download card rather than an empty iframe.
- Non-PDF document bubbles keep their original one-click `<a :download>`; only PDFs route through the lightbox. Applied to the template-header variant too, which had the same regression. They're still in `viewableMedia`, so the arrows remain continuous across the conversation's attachments.

Verification on the updated branch: `vue-tsc` shows only the pre-existing `AccountDetailView.vue` error, `eslint` is clean on `MediaViewerDialog.vue` and reports only the four pre-existing `no-empty` hits in `ChatView.vue` (untouched lines), and `vite build` passes. Re-tested by hand: zoom holds and the PDF no longer reloads while receipts stream in, the ✕ is visible in both themes, and `.docx`/`.xlsx` download in one click again.

